PR #3, opened 2019-09-17
Add travis build

r-ash:
This PR will

* Set up travis build
* Add README and NEWS

Do you want to check the README and add any more details about the package itself as you see fit? I've also bumped the version number as part of this to `0.0.1` with a news item as an example.


jeffeaton:
This is lovely, thanks very much.  I've merged this now and will update the README later.  Later we will also likely remove the INLA dependency.

PR #4, opened 2019-09-18
Add code coverage

r-ash:
Adds travis config for running code cov after a build.


jeffeaton:
I might unmerge this later if I'm feeling really bad about myself.

PR #9, opened 2019-09-27
Add function to return metadata for plotting

r-ash:
This PR will

* Add files containing metadata about locating indicator values in input and output metadata and about colour scales by indicator and country
* Add a function to expose country specific metadata 

I'm expecting that maintenance of this will be done by adding rows to the csv files. I've added a README describing these and explaining how to do some of the expected changes.

At the moment this only contains information for Malawi, I'm hoping this will be expanded on by Jeff & Oli. It will error if a country is not found but perhaps returning some sensible defaults would be better? Again I think that is probably something to ask Jeff & Oli about.

In the hintr API endpoint I will convert this data frame to agreed schema for returning to the front end


R/metadata.R
```diff
@@ -0,0 +1,43 @@
...
+#'
+#' @param country Country to get scale for
+#'
+#' @return List of scale information including colour as a d3 scale chromatic
+#' function name, whether to invert the scale and a min and max value for the
+#' scale.
+#'
+#' @keywords internal
+get_colour_scale <- function(country) {
+  scales <- naomi_read_csv(system.file("extdata", "meta", "colour_scales.csv",
```
richfitz: I would recommend at least adding a `mustWork = TRUE` here (`system.file` has a terrible default) but perhaps also adding a wrapper function for this if you think you'll go past the 2 that are here

R/metadata.R
```diff
@@ -0,0 +1,43 @@
...
+                           package = "naomi"))
+  data <- scales[tolower(scales$country) == tolower(country), ]
+  if (nrow(data) == 0) {
+    stop(sprintf(
+      "Can't retrieve colour scale for country %s. Country not found in configuration.",
```
richfitz: Does colour scale choice belong in this package or in the `hintr` package? It feels more like the latter, or is it used by naomi as well? Is this just so that jeff can change the colour directly (ish)

R/utils.R
```diff
@@ -1,3 +1,7 @@
 naomi_write_csv <- function(...) {
   write.csv(..., row.names = FALSE, na = "")
 }
+
+naomi_read_csv <- function(...) {
```
richfitz: why do these have a `naomi_` prefix?

inst/extdata/meta/colour_scales.csv
```diff
@@ -0,0 +1,8 @@
+country,indicator,colour,min,max,invert_scale
```
richfitz: where are these scales (interpolateViridis, interpolateMagma, etc) actually living? What is the full set of possibilities?

inst/extdata/meta/README.md
```diff
@@ -0,0 +1,34 @@
...
+min - The minimum value to scale the scheme to
+max - The maximum value to scale the scheme to
+invert_scale - Whether to invert the colour scale or not TRUE or FALSE
+
+## FAQ
+
+### How do I add a plot for a new indicator?
+Add a new row to `metadata.csv` which describes what input data type or output data the idicator should be plotted for, the type of plot to use for the indicator and how to locate the relevant data within the dataset.
+
+### How can I change the colour scheme for an indicator for a country?
```
richfitz: related to the comment on colour_scales.csv - what is the full set of possible values here? looks like we need a link to the d3 docs and that is where the actual definition is?

inst/extdata/meta/README.md
```diff
@@ -0,0 +1,34 @@
...
+Edit the `colour_scales.csv` file making sure that the colour column value matches the name of one of the d3-scale-chromatic schemes.
+
+### How can I add a new plot type?
+Add a new row to `metadata.csv` with the correct data type and the new plot type wanted. Then need to tell a developer on the front end that that has been added so they can add front end code to support the new kind of plot.
```
richfitz: can we document the enums in this table too?

inst/extdata/meta/metadata.csv
```diff
@@ -0,0 +1,8 @@
+data_type,plot_type,indicator,value_column,indicator_column,indicator_value,name
```
richfitz: Can you add tests for the data in these csvs so that we confirm that wherever an enum is being used that all values correspond to known values? This will affect pretty much all the non-numeric columns.  If numeric data can be validated too that'd be great.

inst/extdata/meta/colour_scales.csv
```diff
@@ -0,0 +1,8 @@
...
+Malawi,current_art,interpolateViridis,0,80000,FALSE
+Malawi,prevalence,interpolateMagma,0,0.5,TRUE
+Malawi,art_coverage,interpolateViridis,0,1,FALSE
+Malawi,prev,interpolateMagma,0,1,TRUE
+Malawi,artcov,interpolateViridis,0,1,FALSE
```
richfitz: instead of having an invert_scale you could just allow max < min

R/metadata.R
r-ash: I always forget about this, I've fixed that now.

R/metadata.R
r-ash: It is mostly here to Jeff can change the colour but also so Jeff can use the same colour scheme when creating R plots locally. So when users come to download the model outputs they plots look somewhat similar.

R/utils.R
r-ash: Other parts of the package are relying on `read_csv` from `readr` package so I am trying to distinguish between those.

R/metadata.R
richfitz: Great, that seems very reasonable

inst/extdata/meta/colour_scales.csv
r-ash: Scales are from https://github.com/d3/d3-scale-chromatic#api-reference

inst/extdata/meta/README.md
r-ash: I notice the link I included is just to the repo, I will make it to the relevant part of the README. Will that be clear enough you think? There is no block of code that defines the possible values as far as I can see.

inst/extdata/meta/colour_scales.csv
r-ash: Spoke to Emma to see what would be easier for them to use. She says they already have some logic that relies on max < min so I think they would prefer having a boolean flag for inversion.

richfitz: ok, in that case you should enforce min < max in a test here (you could alternatively allow min > max and then in your wrapper function produce the format Emma expects).  But there is a degree of freedom worth removing here

inst/extdata/meta/metadata.csv
r-ash: Okay I've put some checks in. We're now checking that
* Columns which are enums have values from the expected values
* min and max columns are numeric and min >= 0
* invert_scale is logical
* colour is a character
* there are no NULL, NA or empty cells except in `indicator_column` and `indicator_value` columns (which can be empty)

inst/extdata/meta/README.md
r-ash: Added notes about that

PR #10, opened 2019-09-30
Survey vls

jeffeaton:
* Fix error in selecting ART number (and added a test)
* Created a single function `select_naomi_data()` to select data inputs and create a `naomi_data` object. Should be directly callable by front end (see naomi-workflow vignette).
* Implemented support for survey VLS data if a survey has VLS but not ART coverage.

VLS and ART coverage should not be used from the same survey.  This is checked by `select_naomi_data()` function.


tests/testthat/test-model-frame.R
```diff
@@ -6,3 +6,33 @@ test_that("get_age_group_id_out() returns expected groups", {
   expect_equal(get_age_group_id_out(1:17), 1:29)
   expect_equal(get_age_group_id_out(4:17), c(4:21, 25:29))
 })
+
+
+areas <- create_areas(mwi_area_levels, mwi_area_hierarchy, mwi_area_boundaries)
```
r-ash: Can you move lines 11-20 inclusive into a helper file instead? `tests/testthat/helper-model-frame.R` it gets evaluated before the tests so `areas`, `spec`, `naomi_mf` will be available to use within the test code as you have here.

R/model.R
```diff
@@ -495,10 +558,19 @@ anc_testing_artcov_mf <- function(quarter_ids, anc_testing, naomi_mf) {
 
 #' @rdname anc_testing_prev_mf
```
r-ash: Can you update the documentation for `anc_testing_prev_mf` to have an `@param` section for `quarter_id` instead of `quarter_ids`.

tests/testthat/test-model-frame.R
jeffeaton: commit 46f5be6

R/model.R
```diff
@@ -298,13 +302,54 @@ naomi_model_frame <- function(areas,
...
+            logit_nu_mean = logit_nu_mean,
+            logit_nu_sd = logit_nu_sd,  
             M = M,
             Q = Q)
 
   class(v) <- "naomi_mf"
   v
 }
 
+#' @export
```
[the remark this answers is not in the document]
jeffeaton: commit 749d8aa

PR #14, opened 2019-10-18
Iterate model run options

r-ash:
This PR will

* Separate out model run options into section by data type
* Make returning ANC and ART options optional


r-ash:
> Actually, sorry, there is an example that needs updating, or add defaults for these arguments:
> 
> ```
> > get_model_options_template()
> Error in get_model_options_template() : 
>   argument "art" is missing, with no default
> Execution halted
> ```

Sorry, forgot to regenerate documentation. That should be fixed now (and the build).

PR #17, opened 2019-10-21
Plotting metadata returns default if country is not configured

r-ash:
This PR will

* Return a default from plotting metadata function if country is not found in configuration

@jeffeaton Can you review this from a min/max and colour scales POV? Would be good to have sensible defaults here

### DIFF
--- a/R/metadata.R
+++ b/R/metadata.R
@@ -11,8 +11,7 @@
 #'
 #' @examples
 #' get_plotting_metadata("Malawi")
-#' get_plotting_metadata()
-get_plotting_metadata <- function(country = "default") {
+get_plotting_metadata <- function(country) {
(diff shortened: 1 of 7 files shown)


R/metadata.R
```diff
@@ -10,28 +11,33 @@
 #'
 #' @examples
 #' get_plotting_metadata("Malawi")
-get_plotting_metadata <- function(country) {
+#' get_plotting_metadata()
+get_plotting_metadata <- function(country = "default") {
```
hillalex: What's the use case in which country name won't be known when calling this function?

r-ash: Good point - there isn't really a use case for doing that. It will always be known - I will take that out.

PR #23, opened 2019-11-04
Expose no of samples option to allow quicker model run

r-ash:
This PR will

* Get no of samples from model run options and pass to `sample_tmb`

We know that the front end will combine the general options and the advanced options into one options list so this is the form we are expecting the front end to return

### DIFF
--- a/R/run-model.R
+++ b/R/run-model.R
@@ -20,7 +20,7 @@
 hintr_run_model <- function(data, options, output_path = tempfile(),
                             spectrum_path = tempfile(fileext = ".zip"),
                             summary_path = tempfile(fileext = ".zip")) {
-
+  INLA:::inla.dynload.workaround()
   progress <- new_progress()
   progress$start("Preparing input data")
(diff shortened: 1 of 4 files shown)


R/run-model.R
```diff
@@ -94,7 +94,7 @@ hintr_run_model <- function(data, options, output_path = tempfile(),
   progress$complete("Fitting the model")
   progress$start("Generating uncertainty ranges")
   progress$print()
-  fit <- sample_tmb(fit)
+  fit <- sample_tmb(fit, nsample = options$no_of_samples)
```
richfitz: is the full set of these options documented somewhere? is that going to be a `hintr` thing?

r-ash: Not documented. It is specified by naomi in the `inst/extdata/metadata/*_options.json` files. I think if we document that we should probably get it generated from the json automatically or else I can see it getting out of date pretty quick as soon as Jeff changes anything.

PR #24, opened 2019-11-11
Return barchart plotting metadata

r-ash:
This PR will

* Add barchart plotting metadata
* Move metadata to `inst/metadata` from `inst/extdata/meta` this is after talking to Jeff he wants to keep `inst/extdata` as a thing he can send people to for input data and he feels having metadata there might confuse people


DESCRIPTION
```diff
@@ -43,6 +43,7 @@ Imports:
 Suggests: 
     covr,
     datamodelr,
+    DiagrammeR,
```
richfitz: what is DiagrammeR doing in this PR?

r-ash: I added that because in this build https://travis-ci.org/mrc-ide/naomi/builds/610261750#L3125 the vignette failed to build it looks like because of missing DiagrammeR package so included it specifically. Not sure what has changed, I was guessing something about one of the packages being used has changed so it doesn't download the package automatically. Though not sure about that, I could try and take it out and see if we get the same failure.

PR #25, opened 2019-11-13
Update programme data inputs

jeffeaton:
Update the format for programme data inputs:

* Human readable inputs (e.g. `age_group` = 00-04, 05-09, 10-14, ...)
* Annual programme data inputs instead of quarterly inputs.
* Human readable outputs specification
* Adds `spectrum_region_code` to area_hieararchy input


R/data-population.R
```diff
@@ -99,20 +119,21 @@ convert_quarter_id <- function(quarter, year) {
 #' @examples
 #' ## Interpolate Malawi population at level 2 (Zone) at two time points
 #' data(mwi_population_agesex)
-#' quarter_ids <- convert_quarter_id(c(1, 3), c(2016, 2018))
+#' quarter_ids <- convert_quarter_id(c(2016, 2018), c(1, 3))
 #' pop_interp <- interpolate_population_agesex(mwi_population_agesex, quarter_ids)
 #'
 #' @export
-interpolate_population_agesex <- function(population_agesex, quarter_ids) {
+interpolate_population_agesex <- function(population_agesex, calendar_quarters) {
```
r-ash: Looks like you have a real bug in this function. It can't locate `quarter_id` in the dplyr calls lower down and the example doesn't work atm - it is expecting calendar quarters but you pass quarter_ids.

jeffeaton: Thanks very much.  Patched by 82dee15.

PR #28, opened 2019-11-21
Add advanced model run options

r-ash:
This PR will

* Add advanced model run options
* Opportunistically add a section to the general run options for save input data in outputs
* Make survey prevalence run option a multiselect

We should only merge this after #27 has been merged as it builds upon it - merging that will make this PR look much nicer too. We should merge this into master when#27 has been merged, atm point to `projection` branch just to make additions a bit clearer.


R/run-model.R
```diff
@@ -99,6 +99,8 @@ hintr_run_model <- function(data, options, output_path = tempfile(),
   progress$complete("Generating uncertainty ranges")
   progress$start("Preparing outputs")
   progress$print()
+  ## TODO: Include input data in output package based on model options
```
richfitz: does this interact with #30?

inst/metadata/advanced_run_options.json
```diff
@@ -0,0 +1,66 @@
...
+        }
+      ]
+    },
+    {
+      "label": "Number of simulations",
+      "controls": [
+        {
+          "name": "no_of_samples",
+          "type": "number",
+          "value": 3000,
```
richfitz: How long will 3000 take? Is this a sensible default?

r-ash: This is the default used in shiny90, I'm expecting @jeffeaton to review this list and make amendments.

R/run-model.R
r-ash: I think a separate thing - Jeff wanted a flag users could tick to download all the input data along with their outputs. I'm expecting @jeffeaton to add some code which will include the input data in that download based on this new flag. 

Perhaps I should just add a ticket for that and do it to save Jeff some time which he can spend on the model instead.

r-ash:
> Can you clarify for me how the return leg of this works - I can't see it in [mrc-ide/hintr#80](https://github.com/mrc-ide/hintr/pull/80) or here, but we must pass these through to the model run? Or is that to be done elsewhere, and if so in what ticket?

The return is all done by the front end. The front end stores all the user input options as a key value list where the key is the `name` section from the control. So the front end will call the model submit with the all the options which then get passed by hintr to naomi.

richfitz:
Arh sorry about the close/open mashed the wrong button.

> So the front end will call the model submit with the all the options which then get passed by hintr to naomi.

I see that, I just don't see how it makes it through into naomi (i.e., how the number of iterations is altered)

PR #30, opened 2019-11-22
MRC-760: Add manifest and metadata to outputs

richfitz:
This PR adds metadata to the zip bundle produced by naomi.

* The naming scheme I have chosen is: `naomi/{inputs.csv,options.yml,packages.csv}` - @jeffeaton: please let me know if these are not suitable.
* The inputs are stored with role/filename/hash though I think that from the app the filename will actually be the hash by the time it comes through the file store. We can deal with that later and the change would be in hintr or in the hint app
* I have chosen yaml for the options because they are not naturally tabular, this will be forward compatible, and it is relatively human readable. There is no function for reading these back in and it's not hugely likely that a naive roundtrip will work without some care
* The package version information is pretty basic (name, version, base/other/loaded - following `sessionInfo`)
* These are not included in the package_output yet because some additional work would be required to pass the metadata through everywhere. This can be done later IMO as the principle concern here is that we know what was used in the app


jeffeaton:
This is lovely, thank you.

I have suggested one change -- to change the subfolder from `naomi/` to `info/`.

Two more comments/questions:

* Would it be also helpful for us to save the platform and locale information that is output by `sessionInfo()`?  Obviously not an issue for running on our servers, but might be helpful if people use the R package locally.

* It might be preferred to save the hash of the `.DP` and `.PJN` file rather than the hash of the whole `.PJNZ`. But non-essential, and I suggest that we merge this and open a new ticket for any further change on that.

Thanks,
Jeff

richfitz:
Logged the new issues as mrc-1138 and mrc-1139

PR #31, opened 2019-11-22
Add skeleton of model options validation

r-ash:
This PR will

* Add a validate model options function which will be called by hintr to validate options against input data


R/model-options.R
```diff
@@ -33,3 +33,18 @@ get_model_options_template <- function(art, anc) {
...
+#' This validates that a set of model options can be used to run the model
+#'
+#' @param data The set of input data for the model run
+#' @param options Key-value list of model options
+#'
+#' @return TRUE if valid otherwise throw an error
+#' @export
+validate_model_options <- function(data, options) {
+  ## This must return TRUE if valid - otherwise throw an error
+  ## TODO: Add real validation
```
richfitz: :joy: is there a ticket you can point this at?

r-ash: I can add that - I'm expecting Jeff to add this as he discovers problems. Perhaps even a process during workshop as we learn what options cause problems. So I can add a task but I'm not expecting us to sit down and impl this as a whole feature in one go.

r-ash: I've added the ticket number to this

PR #33, opened 2019-11-26
Add hex sticker to README

r-ash:


richfitz:
@r-ash - try exporting as a .png and uploading that

PR #38, opened 2019-11-29
Age groups

r-ash:
This adds a helper function `hintr` needs to get the 5 year age groups


tests/testthat/test-metadata.R
```diff
@@ -107,3 +107,9 @@ test_that("metadata is well formed", {
       all(non_empty(meta[, column]))
     }, logical(1))))
 })
+
+test_that("can get 5 year age groups", {
+  age_groups <- get_five_year_age_groups()
+  expect_length(age_groups, 17)
+  expect_equal(age_groups[[1]], "00-04")
```
jeffeaton: `age_groups` is a vector rather than a list, so canonical way to reference is `[` rather than `[[`

PR #70, opened 2019-12-11
Add translations for naomi strings

r-ash:
This PR will
* Translate strings returned to user
* Translate the names in metadata csv

Coming
* Translate model options page


R/metadata.R
```diff
@@ -51,7 +51,9 @@ get_colour_scale <- function(iso3 = "default") {
 #' @examples
 #' get_metadata()
 get_metadata <- function() {
-  naomi_read_csv(system_file("metadata", "metadata.csv"))
+  data <- naomi_read_csv(system_file("metadata", "metadata.csv"))
```
richfitz: What do you think about doing the translation here by just translating that column?  So do

```
data$name <- vcapply(data$name, t_)
```

and drop the `t_()` that exists in the column

PR #73, opened 2019-12-17
mrc-1263: suppress column type messages from readr

richfitz:
This PR prevents the printing of things like

```
Parsed with column specification:
cols(
  area_id = col_character(),
  sex = col_character(),
  age_group = col_character(),
  year = col_double(),
  current_art = col_double()
)
Parsed with column specification:
cols(
  area_id = col_character(),
  age_group = col_character(),
  year = col_double(),
  anc_clients = col_double(),
  ancrt_hiv_status = col_double(),
  ancrt_known_pos = col_double(),
  ancrt_already_art = col_double(),
  ancrt_tested = col_double(),
  ancrt_test_pos = col_double()
)
```

it does not change the behaviour of readr - it still guesses the column type. It might be nice to do this properly as we know the types of these columns already, but that might have some knock-on effects for validation of user data that I don't anticipate.


jeffeaton:
Thanks for doing this. Yes, this has been on the to-do list to parse specific types. I've updated to add `read_***()` functions for each of the datasets.

These still need validation added. Probably the best way to do this is to copy [validation_asserts.R](https://github.com/mrc-ide/hintr/blob/master/R/validation_asserts.R) from hintr, but should discuss w/ Rob first.  I've put `TODO` comments in the code for this.

Thanks,
Jeff

jeffeaton:
Argh, sorry about that. Done now pending checks.

PR #74, opened 2019-12-18
check that Time 2 is after Time 1

jeffeaton:
This prevents the Kenya error where quarter1 = quarter2 resulting in a self-join --> massive data frame --> using all of the memory and bringing the server down.

Addressed in two ways:

1. Model options check that time 2 is after time 1 and return a nice error message if not.
2. The joins are implemented in a better way so that it wouldn't explode the memory.


richfitz:
In general, please pull to get an up-to-date master before starting a new branch to save the integration work as above - this is particularly important after having extracted strings for translations as if you're working against old sources there will be many conflicts (as above).  

In addition, version 42 has already been incorporated into hintr, so I've increased the version number here. Please update your sources on this branch.

jeffeaton:
Thanks for that -- sorry.  I was doing this in flight, but indeed also hadn't realized I didn't have the current master pulled.

Re the translations, is there guidance written about the proper way to add strings now?  I think I can figure it out from the patterns in the commit, but thought I'd check in case any guidance I should follow.

Thanks,
Jeff

richfitz:
No worries. Add newly added strings to mrc-1267 please, but don't worry about moving them into the translation dictionaries yet (and in particular please don't move them for cases that do not have a test associated). I have to pop out to run an errand but will check integration with hintr in the afternoon

PR #78, opened 2020-01-07
Add default value sections to options template

r-ash:


jeffeaton:
Brilliant, thanks. Is it also possible to add a default value for `calendar_quarter_t1`, which may or may not be provided?

Per workshop feedback, I am going to add an optional `midpoint_calendar_quarter` column to the `survey_indicators` dataset, which would be used to populate this.

r-ash:
> Brilliant, thanks. Is it also possible to add a default value for `calendar_quarter_t1`, which may or may not be provided?
> 
> Per workshop feedback, I am going to add an optional `midpoint_calendar_quarter` column to the `survey_indicators` dataset, which would be used to populate this.

Just made a small change to this. The calendar quarter t2 default is now in the configuration (instead of being set by hintr - there will be a corresponding hintr PR coming your way soon for the other half to this one)

For calendar quarter t1 default, if that column is not populated we should fall back to the default being empty? Not sure how that will work but can check what that would look like. Though we should probably do that in a separate ticket when you add that column.

PR #80, opened 2020-01-08
Add INLA and tmbstan model fitting

jeffeaton:
This PR adds support to do model fitting using alternative inference strategies: INLA and TMB-Stan.

These fitting approaches are supported for methodological model testing and comparison only, not any deployment. Thus this PR should be invisible to user.

There is a refactor to `hintr_run_model()` to move much of the work to a new function `naomi_prepare_data()`. This doesn't change any of the functionality of `hintr_run_model()`, but perhaps have a quick look at that to make sure it doesn't have any unforeseen consequences. 

The tests for `fit_inla()` and `fit_tmbstan()` are a bit time consuming.  We might want to disable those by default now or later.

Thanks,
Jeff


jeffeaton:
Huzzah -- `skip_on_covr()` for those new heavy tests resolved the Travis build timeout. Thanks for that.

NEWS.md
```diff
@@ -11,6 +11,13 @@
 
 * Add read_***() functions with col_type parsers.
 
+# naomi 0.0.46
```
r-ash: The news item is for naomi v 0.0.46 but the current version in the Description is 0.0.49?

Can you bump the version number to 0.0.50 and set the news item as that?

PR #82, opened 2020-01-14
approximate design effect to adjust survey effective sample size

jeffeaton:
This PR adds three numeric inputs under the 'Advanced Options' tab.  This needs review in `hintr`.  Ideally new strings also need translated, but not the end of the world if not.

Thanks,
Jeff


inst/metadata/advanced_run_options.json
```diff
@@ -1,5 +1,5 @@
-{
-	"label": "t_(OPTIONS_CALIBRATION_LABEL)",
+t": "{
+	"label_(OPTIONS_CALIBRATION_LABEL)",
```
r-ash: Oops - an accidental highlight and drag? This is the wrong syntax.

inst/metadata/advanced_run_options.json
```diff
@@ -214,6 +214,36 @@
 					"helpText": "t_(OPTIONS_ADVANCED_LOGIT_HELP)"
 				}
 			]
+		},
+	        {
+			"label": "Approximate district survey design effect",
```
r-ash: Can you move all the strings into the resource bundle https://github.com/mrc-ide/naomi/blob/master/inst/traduire/en-translation.json#L78

This is for each string which gets returned to the user (and so requires translation) pick a key name and use that here e.g.
```
"label": "t_(OPTIONS_DESIGN_EFFECT_LABEL)"
```

Then there should be a line in the resource bundle
```
"OPTIONS_DESIGN_EFFECT_LABEL": "Approximate district survey design effect"
```

That should be for all user facing strings here

inst/metadata/advanced_run_options.json
```diff
@@ -214,6 +214,36 @@
...
+                       "helpText": "Approximate design effect to scale effective sample size for survey observations. Typically this will be smaller than the full survey design effect.",
+		        "controls": [
+				{
+					"name": "deff_prev",
+					"label": "Prevalence",
+					"type": "number",
+					"required": true,
+				        "value" = 1.0
```
r-ash: Should be 
```
"value": 1.0
```

jeffeaton: Sigh, thanks. This makes a strong case for just going to bed.

Fixed in 75406ff

inst/metadata/advanced_run_options.json
jeffeaton: Thanks. Done in 2efcab9.

At some point I thought Rich mentioned something like that everything that goes into the translation resource needs to be tested. Do I need to do anything further on this?

r-ash: I don't know about that - everything there needs to be translated but I don't believe we're testing every string at the moment. Maybe @richfitz can comment? 

This looks fine to me - I will ask Anne to give us a translation of these strings and add that to this PR before merging.

inst/metadata/advanced_run_options.json
```diff
@@ -214,6 +214,36 @@
 					"helpText": "t_(OPTIONS_ADVANCED_LOGIT_HELP)"
 				}
 			]
+		},
+	        {
+		       "label": "t_(OPTIONS_ADVANCED_APPROX_DEFF_LABEL)",
+                       "helpText": "t_(OPTIONS_ADVANCED_APPROX_DEFF_LABEL)",
```
r-ash: Sorry another thing came up after putting this through our schema. You can't have help text here. Help text goes on the individual controls only.

jeffeaton: Thanks -- removed in 7144474. (Also that is a pretty nice commit hash.)

PR #89, opened 2020-01-14
remove NA rows in read_*() functions

jeffeaton:
Brings behaviour of naomi in line with hintr #139 https://github.com/mrc-ide/hintr/pull/139

### DIFF
--- a/R/read-data.R
+++ b/R/read-data.R
@@ -22,6 +22,8 @@ read_population <- function(file) {
   val <- read_csv_partial_cols(file, col_types = col_spec)
   readr::stop_for_problems(val)
 
+  val <- drop_na_rows(val)
+
   missing_cols <- setdiff(required_cols, names(val))
   if(length(missing_cols))
     stop(paste0("Required columns not found: ", paste(missing_cols, collapse = ", ")))
@@ -58,6 +60,8 @@ read_survey_indicators <- function(file) {
   val <- read_csv_partial_cols(file, col_types = col_spec)
   readr::stop_for_problems(val)
 
+  val <- drop_na_rows(val)
+  
   missing_cols <- setdiff(required_cols, names(val))
   if(length(missing_cols))
     stop(paste0("Required columns not found: ", paste(missing_cols, collapse = ", ")))
@@ -88,6 +92,8 @@ read_art_number <- function(file) {
   
   val <- read_csv_partial_cols(file, col_types = col_spec)
   readr::stop_for_problems(val)
+
+  val <- drop_na_rows(val)
 
   missing_cols <- setdiff(required_cols, names(val))
   if(length(missing_cols))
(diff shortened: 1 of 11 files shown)


R/read-data.R
```diff
@@ -22,6 +22,8 @@ read_population <- function(file) {
   val <- read_csv_partial_cols(file, col_types = col_spec)
   readr::stop_for_problems(val)
 
+  val <- drop_na_rows(val)
```
r-ash: Just looking at this there is quite a lot of code duplication. I'd be tempted to move 23-29, 61-67, 94-100 and 149-155 into `read_csv_partial_cols`

R/read-data.R
```diff
@@ -185,3 +193,8 @@ read_area_merged <- function(file) {
 read_csv_partial_cols <- function(...){
   suppress_one_warning(readr::read_csv(...), "The following named parsers don't match the column names")
 }
+
+drop_na_rows <- function(x) {
+  na_rows <- apply(is.na(x) | x == "", 1, all)
```
r-ash: Looking at the way this is used x should never be "" here as `readr::read_csv` makes any empty string an NA. I think not a problem to leave it as is though.

R/read-data.R
jeffeaton: I was also tempted to do this when implementing. On the other hand, if felt like `read_csv_partial_col()` does 'a thing', and wasn't sure about extending it.  But if you and I had the same instinct...should do it...

R/read-data.R
jeffeaton: I noticed that when I was writing the test.  This behaviour mimics hintr: https://github.com/mrc-ide/hintr/blob/master/R/read_data.R#L24   

Is there any other reason that hintr checks the "" case?

r-ash: Yeah - in hintr we're using `read.csv` from R utils instead of `readr::read_csv`. The utils `read.csv`  will use "" if it thinks the column type is a character.

```
t <- tempfile()
writeLines(c("string_col,number_col", "testing,2", ","), t)
read.csv(t)
```

Gives
```
  string_col number_col
1    testing          2
2                    NA
```

jeffeaton:
Got it, thanks.

On 16 Jan 2020, at 09:40, Rob <notifications@github.com<mailto:notifications@github.com>> wrote:


@r-ash commented on this pull request.

________________________________

In R/read-data.R<https://github.com/mrc-ide/naomi/pull/89#discussion_r367317930>:

> @@ -185,3 +193,8 @@ read_area_merged <- function(file) {
 read_csv_partial_cols <- function(...){
   suppress_one_warning(readr::read_csv(...), "The following named parsers don't match the column names")
 }
+
+drop_na_rows <- function(x) {
+  na_rows <- apply(is.na(x) | x == "", 1, all)


Yeah - in hintr we're using read.csv from R utils instead of readr::read_csv. The utils read.csv will use "" if it thinks the column type is a character.

t <- tempfile()
writeLines(c("string_col,number_col", "testing,2", ","), t)
read.csv(t)


Gives

  string_col number_col
1    testing          2
2                    NA

PR #90, opened 2020-01-16
add time 3 projection

jeffeaton:
This PR adds a projection to a future time point (September 2020 by default) for planning and target setting needs.

Key changes: 

* Outputs appear in 'indicators.csv' file for a third calendar quarter.
* Adds a text field to the 'Output options' section of the model options page to specify the calendar quarter for the projection, default "CY2020Q3".

Ideally the projection calendar quarter should be populated as a select control with labels similar to the survey and current point point selects, rather than as the raw calendar_quarter string. However, essentially nobody should need to change this from the default value "CY2020Q3" and it doesn't seem timely to try to add another endpoint for pre-populating this right now.

Thanks,
Jeff


jeffeaton:
I have updated this now so that the the model option is a 'select' control, consistent with the Time 1 and Time 2 selection controls.

There are also several additional commits adding:

* traduire hooks for calendar quarter labels.
* traduire hooks for indicator labels and indicator descriptions.
* tests for traduire hooks in model outputs.

@r-ash I think this is ready for your review now.

There are also several strings missing french translations. @richfitz -- are these strings you are able to do or should we send them to someone?

r-ash:
Looks good to me - I think we should try and get the translations in before merging this.

richfitz:
I'll take a look - there were others that needed looking at too @r-ash? I can do a quick hack and then check later perhaps

r-ash:
> I'll take a look - there were others that needed looking at too @r-ash? I can do a quick hack and then check later perhaps

The list here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1267 *should* be complete
We need to go through and review app afterwards to double check that. I expect there are some things we're missing.

richfitz:
thanks - lots of those are easy. Could you give the linter a quick spin and see if it detects them?

r-ash:
lintr looks good - it picks up the strings which aren't in the french resources at all. Several of them are just in the metadata json which we wouldn't expect it to highlight. Looking good.

richfitz:
> One comment so far - rest coming later

I can't see the actual comment I (tried to) left yesterday anywhere

PR #99, opened 2020-01-21
Fix test model options

jeffeaton:
This affects only naomi tests and is super tiny. 

Thanks,
Jeff


richfitz:
Looks like travis has succeeded with this, but the hooks back to gh have failed perhaps?

PR #104, opened 2020-01-24
mrc-1320: better error messages in code mismatch

richfitz:


richfitz:
No tests for now, checked locally

PR #110, opened 2020-03-20
mrc-1436: Model can be run taking SessionFile formatted data

r-ash:
This PR will
* Update `naomi_model_run` so that it can be called either with a list of files or with a list of SessionFile formatted data

Where a SessionFile is a list with properties path, hash and filename. This is so we can get the filename as uploaded from the front end and pass it through to the downloaded spectrum digest.

Decided to keep the support for calling with a list of files as I think this would be a real pain when working with programmatically if you have to construct the list of SessionFile for the input data.


r-ash:
Have confirmed this is passing tests within docker container

R/run-model.R
```diff
@@ -320,3 +336,35 @@ naomi_info <- function(data, options) {
...
+##     filename = "filename"
+##   ),
+##   input2 = list(
+##     ...
+##   ),
+##   ...
+## )
+## From either a list of file paths or returns passed in data unchanged
+format_data_input <- function(data) {
+  if (all(vapply(data, is.character, logical(1)))) {
```
jeffeaton: Is correct reading of this that it requires all to be either 'path' format or all 'list' format?  If so, should this check be applied to each file separately?

I don't think this is a necessary change right now -- there's not a clear use case for it at present.

r-ash: That's right - all either list or path. Mostly because I agree I can't see a use case for having a mix. Would be fairly simple to add if we end up having a mix sometime in the future.

PR #111, opened 2020-03-20
Return area scope in model result

r-ash:


r-ash:
> This looks good to me. Will the file name also include time stamp per request from UNAIDS? I think that's probably something that can be affixed by hintr and not needed in here.

Yep gets added in hintr. Thanks!

PR #118, opened 2020-06-25
Use naomi base image

r-ash:


docker/Dockerfile
```diff
@@ -1,69 +1,11 @@
...
-        libssh-dev \
-        libudunits2-dev \
-        libv8-dev \
-        libxml2-dev \
-        protobuf-compiler \
-        texinfo \
-        texlive \
-        texlive-fonts-extra \
-        zlib1g-dev
+FROM mrcide/naomi_base:latest
```
richfitz, line 1: I should have noticed this before sorry. I think that docker tags are almost always dashes - PR for the actual image https://github.com/mrc-ide/naomi.docker/pull/2 and then this can become naomi-base

docker/Dockerfile
```diff
@@ -1,69 +1,11 @@
...
-    tidyr \
-    tidyverse \
-    tmbstan \
-    traduire \
-    viridis \
-    withr \
-    yaml \
-    zip \
-    zoo
+COPY docker/Rprofile.site /usr/local/lib/R/etc/Rprofile.site
```
richfitz, line 4: You don't need to do this if it's in the base image :)

PR #119, opened 2020-06-26
MRC-1666: Remove INLA and tmbstan models

r-ash:
* Move INLA and tmbstan models into naomi.extensions see https://github.com/mrc-ide/naomi.extensions/pull/1

This sets the verison to 1.0.0 too :tada: as we're using this in production I think it is about time we are at 1.0.0. 

Also want to mention some of the other functions which aren't called by the API at any point. Some of these we might be able to remove others we might be able to move somewhere else. But I expect there is an argument for leaving some in.

`data-raw` there are several functions in this group which are used to process the data in `data-raw` but are never actually used by the package. Might be a candidate for moving out but I think most of these won't be a big deal if they stay in.
* `calc_survey_hiv_indicators`
* `expand_survey_clusters`
* `spread_areas`
* `th_map`
* `compare_boundaries`
* `check_boundaries`
* `get_mid_calendar_quarter`
* `cmc_date`
* `gather_areas`

Never called - I wonder if we can just remove some of these?
* `read_sf_zip_list`
* `creade_edge_list`
* `st_read_zip` - in data-areas but never called
* `report_tmb`
* `get_area_collection` - never called (commented out calls in some vignettes)

Others which aren't called by the API but should probably stay in
* `save_output_package` - used in examples v small in scope
* `export_datapack` - never called but seems to be closely related to package code and not a huge gain from moving it
* `subset_output_package` - similar as above 
* `read_output_package` - similar again


r-ash:
After discussing

At some point we might move the data-areas related function to a new package that does data prep. That package will include shared metadata with ADR and naomi so naomi will depend on it. 

We've removed some functions but `get_area_collection` needs some reworking at some point. Jeff wants to keep in `creade_edge_list`, `report_tmb`.

jeffeaton:
🌺

PR #120, opened 2020-08-06
Update metadata to include uncertainty for all output data

r-ash:


inst/metadata/metadata.csv
```diff
@@ -6,14 +6,14 @@ survey,choropleth,recent,est,,,indicator,recent,t_(PROPORTION_RECENTLY_INFECTED)
 survey,choropleth,vls,est,,,indicator,vls,t_(VLS)
```
Nicolas-Dolan, line 6: Potentially a dumb question but does the SAP component not have uncertainty ranges then? Are they only generated for the model output?

EmmaLRussell: @Nicolas-Dolan Yeah, I think all the input data is pure concrete, gathered numbers - they won't be completely accurate of course, but we can only apply uncertainty to the values coming out of the model. So the table component is going to have to deal with uncertainty metadata not being provided, in which case just don't show the error min and error max columns.

PR #124, opened 2020-09-28
MRC-1790: Rename summary download to coarse-output

r-ash:
This is in preparation for adding an endpoint for summary report download to avoid confusion between the two endpoints. It makes this endpoint more representative of what it is actually doing.

This shouldn't be merged until there are PRs all the way to the front end which ensure that this switch works - it is fine to be sat around as a PR for a while

See also https://github.com/mrc-ide/hintr/pull/195


hillalex:
https://github.com/mrc-ide/hint/pull/346 now merged

richfitz (approved):

PR #125, opened 2020-09-29
MRC-1792: Add model calibration options

r-ash:
This PR will
* Add a function for returning model calibration options

Note that this PR dupes calibration options in model run options and model calibration options. The PR to remove options from the model run section I will add separately as removing those is blocked on updating the front end


inst/metadata/calibration_run_options.json
```diff
@@ -0,0 +1,165 @@
+{
```
richfitz, line 1: I feel we use spaces elsewhere for indentation - might be worth updating this so we don't end up with a weird mix of tabs and spaces

richfitz (approved):
This looks fine to me, so long as @jeffeaton is happy with the result. I'd probably un-tabbify that json file though

inst/metadata/calibration_run_options.json
r-ash: Thanks, hadn't noticed that but I've switched all tabs in options to 2 spaces.

r-ash:
Going to merge this as is - the options can be updated and finalised whenever is convenient

inst/metadata/calibration_run_options.json
jeffeaton: I think the main culprit here is my Emacs config defaulting to tabs and sneaking them in there!! I need to figure that out.

jeffeaton (commented):
This looks great to me -- thanks very much.

PR #129, opened 2020-10-01
MRC-1796: Add separate model result calibration function

r-ash:
This PR should
* export `hintr_calibrate` to calibrate model run with list of options
* Not break the existing `hintr_run_model` for the front end i.e. we need to still return take calibration options in model run and return a result in the same format

Need to update hintr PR to work with this before merging but it shouldn't be blocked on front end


richfitz (approved):

jeffeaton (approved):
Thanks for doing this work.  As discussed, saving the `calibration_data` as an additional output is a good solution to keep this moving forward. But the ideal solution is for the output package to contain all of the information needed to uncalibrate and recalibrate.  

Let's merge this now to let the interface move forward and create another ticket to refactor this within naomi package later.

PR #133, opened 2020-10-14
catch segfault when change computer/num threads

kklot:
not sure if naomi ever needs to do this, if there is a time when we save the fit from one computer and redo the sampling in another without refitting then this would be needed.


jeffeaton:
Thanks, this is great! I've confronted this a few times, but never figured out if there's a way to fix. Can you explain a bit more about what the code is doing?

Also can you please:
* Increment the version number
* Add an entry to NEWS.md

Usually would also request a regression test reproducing the error and confirming fixed. But I think testing for segfaults is hard...

Thanks,
Jeff

kklot:
TMB works with external pointers (to data, ADFunc, ADHess..), when we change data/computer/number of threads TMB needs to update those otherwise they can be null pointer. This check one of the main pointers, if it's null we update the pointer with TMB's `retape()` but keep the current values (e.g. `last.par`) not resetting them to default.

jeffeaton (approved):
Thanks for this -- great.

PR #140, opened 2020-10-19
add basic hintr-example run vignette

jeffeaton:
This creates a basic vignette to emulate a hintr model run.

I tried editing the Makefile to compile the vignettes, but it didn't go well...


r-ash (approved):
Updating other vignette in a different PR sounds fine.

PR #150, opened 2020-11-03
Update summary report with header

rtesra:


inst/report/summary_report.Rmd
```diff
@@ -679,3 +670,6 @@ x4 %>% gt::gt(rowname_col = "area_name") %>%
 
 ```
 
+### **Naomi methods**
+
+[place holder text for Naomi methods]
```
rtesra, line 675: @jeffeaton You can add the methods in here as raw text

R/outputs.R
```diff
@@ -672,18 +672,28 @@ save_output <- function(filename, dir,
...
-  report_path_dir <- normalizePath(dirname(report_path), mustWork = TRUE)
   report_filename <- basename(report_path)
+  report_path_dir <- normalizePath(dirname(report_path), mustWork = TRUE)
   output_zip_path <- normalizePath(output_zip, mustWork = TRUE)
-  rmarkdown::render(rmd_path, params = list(
+  ## Render uses relative paths to locate the html file and pacakge author
+  ## advises against using output_dir see:
+  ## https://github.com/rstudio/rmarkdown/issues/587#issuecomment-168437646
+  ## so set up a temp directory with all report sources and generate from there
+  tmpd <- tempfile()
```
richfitz, line 683: You might use withr here for these , but this is ok too

R/outputs.R
```diff
@@ -672,18 +672,28 @@ save_output <- function(filename, dir,
...
+  file.copy(list.files(system_file("report"), full.names = TRUE), ".")
+
+  rmarkdown::render("summary_report.Rmd", params = list(
     output_zip = output_zip_path),
-    output_dir = report_path_dir,
     output_file = report_filename,
     quiet = quiet
   )
 
+  file.copy(report_filename, report_path_dir)
```
richfitz, line 697: I advise using stopifnot() around file.copy as it does not error on failure. Or use a function from `fs`, which generally has better behaviour.

Do we not need the whole directory?

richfitz (changes requested):
Couple of small comments here

R/outputs.R
r-ash: `withr` is nicer, I have switched that over

R/outputs.R
r-ash: Thanks, switched to use `fs` we only need to copy the single file. The rendering pulls the images & banner html into the output document so it is all available in the single file.

richfitz (approved):

PR #152, opened 2020-11-14
patch summary report for case where area_name not unique

jeffeaton:
Generating the table in the summary report includes an argument `spread(indicator, val)`, where `area_name` is the remaining column as the ID column. The `area_name` is not guaranteed to be unique, and so this will not work.  The `area_id` is guaranteed to be unique and should always be used when manipulating, merging, or filtering data.

I haven't made any edit to the table, so there's an issue that the rows will not be distinguishable if the area name is duplicated. We need to think more about this.

If we are happy staying with an HTML formatted report, one elegant way to handle this might be to add a tool tip showing the full area hierarchy path.

Thanks,
Jeff

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: naomi
 Title: Naomi Model for Subnational HIV Estimates
-Version: 2.0.1
+Version: 2.0.2
 Authors@R: 
     person(given = "Jeff",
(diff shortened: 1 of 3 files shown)


rtesra (approved):

jeffeaton:
This is addressed in #167. Closing this PR.

Issue #155, opened 2020-11-20
Update calibration options text


r-ash:
Now these are split over model options and calibration options we should update the help text

jeffeaton:
Addressed in #162, closing.

PR #156, opened 2020-11-20
Update model options

r-ash:
This PR will
* Split "approximate district survey design" over 3 rows to stop messy looking wrapping of 3rd option
* Switch "district j" to "neighbouring district" in logic prior mean tooltip
* Remove full stops from HIV prevalence & ART coverage tooltips
* Move "Time varying ART attendence" to bottom of ART options

We still need to do #155


jeffeaton:
Looks very good -- thanks very much for doing these.

jeffeaton (approved):
Looks very good. Thanks for doing these.

PR #157, opened 2020-11-22
Add new ANC testing cascade outputs

jeffeaton:
This PR implements updates to the model produce estimates for the ANC testing cascade by district (number of ANC clients, number of HIV positive pregnant women, and number already on ART).

Below are notes describing the new outputs, model changes, and UI changes. The NEWS.md updates are relatively detailed as well. Of particular note for reviewing:
* This PR incorporates PR #164 (which are small cosmetic edits). That should be merged first, or closed and taken together in this PR.
* This PR rolls back some of the changes to the ANC testing data model options section we deployed Friday (sorry).
* Crude translations of new French label strings are added with google translate. Added issue #169 to verify them.
* Should require no changes for hintr or hint.

-----

### New model outputs
The following indicators are output by 5-year age group and aggregated for all three time periods
* Number of ANC clients
* Number of HIV-positive clients
* Number of ANC clients already on ART
* HIV prevalence among ANC attendees
* ART coverage among ANC attendees
* Number of ANC clients tested positive: _calculated as difference of total positive clients minus already on ART._
* Number of ANC clients tested negative: _calculated as difference of total clients minus positive clients._
* Number of ANC clients initiated on ART: _output equivalent to number of clients tested positive._

Multiple outputs reporting the same thing is to have a separate indicator for each Data Pack input. There is a pending query with Scott Jackson whether it is important to make an effort to distinguish 'known positive' versus 'already on ART' clients considering these are poorly distinguished in data and awareness of status is likely underreported.

### Model changes:
* ANC bias parameter is added on age-specific regression instead of after aggregating ANC prevalence in order to produce age-stratified outputs consistent with observed data.
* Age-specific fertility rate ratios for HIV positive women and women already on ART are calculated from PJNZ inputs using EPP-ASM and input as offsets to the ANC prevalence and ANC ART coverage regressions.
* More flexible specification for ANC testing input data: now constructed as aggregation of area/age group predictions. This enables inputting of ANC testing data by coarser regions (e.g. admin 1) or age-stratified if desired (but not recommended yet).
* Likelihood added for observed number of ANC clients at time 2 to calibrate predicted number of clients. Includes accounting for less than full year of data reflected in ANC testing data.
* Additional district random effects to calibrate overall fertility rate to observed number of ANC clients. This should not be interpreted as 'true' fertility, but a catch all for misspecification of fertility rates and patterns of ANC attendance (proportion attending + neighbouring district, which is not explicitly modelled).

### User interface changes:
* Adds model option to select input data about number of ANC clients at the second observation period, and the number of months of data included.
  * This rolls back the change last week to require selecting the number of months of data for each ANC prevalence and ANC ART coverage observation. That information is not needed -- only the number of months for the total number of clients data point, and the interface is cleaner without. Sorry for that.
* Adds several indicators to outputs -> metadata.csv and colourscales.csv.

### Code changes
* Model now involves simulation of the EPP-ASM model to calculate FRR 
* Data extraction for EPP-ASM provides all of the indicators that were formerly ascertained from `specio` with fewer unzips and reads of the .DP file --> remove dependency on `specio`. (Next year to get back to integrating those...)


inst/traduire/en-translation.json
```diff
@@ -116,5 +116,11 @@
...
     "SCOPE_AREAS_MISSING_HIERARCHY_PLURAL": "Scope areas {{missing_areas}} not found in hierarchy.",
     "SHAPE_SPECTRUM_REGION_ALL_NA": "Spectrum region code column is all missing in your shape file. Update the shape file with Spectrum region code for the estimation level.",
-    "INVALID_CALIBRATE_OBJECT": "Can't calibrate this model output please re-run model and try calibration again"
+    "INVALID_CALIBRATE_OBJECT": "Can't calibrate this model output please re-run model and try calibration again",
+    "ANC_CLIENTS": "ANC clients",
+    "ANC_PLHIV": "HIV positive ANC attendees",
+    "ANC_ALREADY_ART": "ANC attendees already on ART",
+    "INDICATOR_LABEL_ANC_CLIENTS": "Total number attending first ANC",
+    "INDICATOR_LABEL_ANC_PLHIV": "Number of HIV positive women attending ANC",
+    "INDICATOR_LABEL_ANC_ALREADY_ART": "Number already on ART prior prior to first ANC"
```
jeffeaton, line 125: Additions to the English dictionary. French translations needed.

jeffeaton: Crude translations added to French dictionary from Google translate. Created issue #169 to check these.

jeffeaton:
This is ready for review. Note that it incorporates #164 (which is small) and overwrites some of those changes, so either merge that first or close and take them together here.

inst/metadata/anc_run_options.json
```diff
@@ -3,19 +3,19 @@
   "description": "t_(OPTIONS_ANC_DESCRIPTION)",
   "controlGroups": [
     {
-      "label": "t_(OPTIONS_ANC_PREVALENCE_LABEL) - t_(OPTIONS_T1_LABEL)",
+      "label": "t_(OPTIONS_ANC_CLIENTS_LABEL)",
```
r-ash, line 6: This looks a bit weird to have a column where first entry is time 2 and then subsequent in column are both time 1. Though not sure how you would neaten it up as first row is required

inst/metadata/anc_run_options.json
```diff
@@ -3,19 +3,19 @@
...
-          "label": "t_(OPTIONS_T1_LABEL)",
+          "name": "anc_clients_year2",
+          "label": "t_(OPTIONS_T2_LABEL)",
           "type": "select",
-          "helpText": "t_(OPTIONS_T1_HELP)",
+          "helpText": "t_(OPTIONS_T2_HELP)",
           "options":
-            <+anc_prevalence_year1_options+>,
+            <+anc_prevalence_year2_options+>,
           "required": false
```
r-ash, line 15: Also intended that this isn't required but the entry in the same row as this is required?

r-ash (changes requested):
Got an error running this with Malawi 0f9ef77efe7174bbd1290f44b462d687 possibly data out of date. Would be good if we can check this as part of validation if that is the issue. Also this seems like a good time to test that you can pull the data from that ID using the endpoint from last year. As far as I'm aware that hasn't been used since last year?

jeffeaton:
> Got an error running this with Malawi 0f9ef77efe7174bbd1290f44b462d687 possibly data out of date. Would be good if we can check this as part of validation if that is the issue. Also this seems like a good time to test that you can pull the data from that ID using the endpoint from last year. As far as I'm aware that hasn't been used since last year?

Regarding the error, the reason for this is because the front end is passing the `anc_clients_year2_num_months` option as a string ("12") and that function where it erred was interpreting it as a numeric.  I've fixed this in 1fccd1f. 

The reason this got through is another case of the model options list in `tests/testthat/setup-run-model.R` not being fully consistent with the JSON schema. I've created an issue to figure out how to write a test against this (requires figuring out how to fill in those glue tags with something): https://github.com/mrc-ide/naomi/issues/171.

inst/metadata/anc_run_options.json
jeffeaton: I agree it looks weird. But the logic is that it's not required to specify a year for calibrating this. But if a year is specified, then the number of months is required to be specified.  This was the only way I could think to encode that?

r-ash:
> > Got an error running this with Malawi 0f9ef77efe7174bbd1290f44b462d687 possibly data out of date. Would be good if we can check this as part of validation if that is the issue. Also this seems like a good time to test that you can pull the data from that ID using the endpoint from last year. As far as I'm aware that hasn't been used since last year?
> 
> Regarding the error, the reason for this is because the front end is passing the `anc_clients_year2_num_months` option as a string ("12") and that function where it erred was interpreting it as a numeric. I've fixed this in [1fccd1f](https://github.com/mrc-ide/naomi/commit/1fccd1f94bc49328bfb6bfa81a3195e948146a21).
> 
> The reason this got through is another case of the model options list in `tests/testthat/setup-run-model.R` not being fully consistent with the JSON schema. I've created an issue to figure out how to write a test against this (requires figuring out how to fill in those glue tags with something): #171.

Thanks, agreed sounds like a good idea to make sure our test options are consistent with schema

jeffeaton:
As discussed -- closing this PR and considering it together with #170 .

PR #158, opened 2020-11-23
Print progress message during model fitting

r-ash:
This PR will
* print progress messages update on time elapsed and iteration during `fit_tmb`, this adds to helpText which should be passed straight through hintr to the endpoint and is what hint is expecting


R/run-model.R
```diff
@@ -113,7 +113,8 @@ hintr_run_model <- function(data, options, output_path = tempfile(),
   fit <- fit_tmb(tmb_inputs,
                  outer_verbose = ifelse(is.null(options$outer_verbose), FALSE, options$outer_verbose),
                  inner_verbose = ifelse(is.null(options$inner_verbose), FALSE, options$inner_verbose),
-                 max_iter = ifelse(is.null(options$max_iterations), 250, options$max_iterations))
+                 max_iter = ifelse(is.null(options$max_iterations), 250, options$max_iterations),
```
richfitz, line 116: while you're changing code around here:

* what is 250, and should that get a variable name
* prefer `if (is.null(options$max_iterations) 250 else options$max_iterations` or `options$max_iterations %||% 250` to `ifelse`

richfitz (changes requested):
This looks great - a suggestion in unrelated code below

R/run-model.R
r-ash: Tidied that, 250 is just default max iterations. Same as the default on `fit_tmb` function itself.

richfitz (approved):

PR #164, opened 2020-11-27
Edits to ANC model options block lables

jeffeaton:
To try to make the UI look a bit better.

Also added into this PR:
* Add `area_name` field to the Malawi example datasets to conform to ADR specifications.


r-ash:
see #157

PR #172, opened 2020-12-02
get_colour_scale() return default colour scales if a custom scale not specified 

jeffeaton:
* Added a test to ensure that a colour scale is returned for all indicators in `indicator_meta.csv` (which fails before patch).
* Updated get_colour_scales() logic to first get default colour scales, then replace any that have custom scales for an ISO3.

The test is not perfect because if someone in future updates "MWI" to have custom scales for all indicators, it will not check the condition. I think we would need to mock something to write a more robust test?

Thanks,
Jeff


jeffeaton:
Nice one

PR #174, opened 2020-12-03
mrc-1969 Makes advanced model options collapsed by default

Nicolas-Dolan:
This will cause the Advanced section in the model options to be collapsed by default. Most users do not use this section so having it collapsed to begin with helps to declutter the page. We might also want to make all sections collapsible. The potential downside is that it might cause the user to lose track of some of the options they had collapsed.


r-ash (approved):
Amazing, thanks Nick

Issue #175, opened 2020-12-03
drop validate_options() from hintr_run_model()?


jeffeaton:
@r-ash: Now that the 'validate' button on the model options page is active, can we remove `validate_model()` from `hintr_run_model()`?

https://github.com/mrc-ide/naomi/blob/1ead78db11e98f83c7ae73982cc45943a86384c3/R/run-model.R#L71

r-ash:
We will need to update the progress messages too which is the only thing which is potentially tricky. It shouldn't be too bad though, biggest faff is probably updating tests of the progress messages.

jeffeaton:
This is completed by #181. Closing issue.

PR #181, opened 2020-12-04
Make running validation of inputs optional in model run

r-ash:
This addresses #175 

I think useful to make this option as still instances (mostly when running programmatically i.e. when generating test data) when I want to be told about validation errors


jeffeaton (changes requested):

jeffeaton:
PR looks good. Only thing for changes is I think comment on number of states should be '4' now instead of 5

jeffeaton (approved):

PR #188, opened 2020-12-08
Model options parsing patches

jeffeaton:
This fixes a few issues identified by @rtesra in testing:

* If a user selects a year for `anc_clients_year2` then unselects it, the front end passes a string "" instead of a NULL. This was already handled for `
* Really nasty one where if no ART data is uploaded, then the ART options block is not loaded and none of the options are provided. so R autocompletes `is.null(options$artattend)` to check `options$artattend_log_gamma_offset`.
* Move ad hoc model options parsing from `hintr_run_model()` into a separate function.


r-ash (approved):
This looks good

PR #190, opened 2020-12-09
Issue 165: Add default model options for 

r-ash:
* Select most recent household survey for prevalence based on survey_mid_calendar_quarter
* Select most recent household survey for art_coverage if data exist
* If ANC data exist in the most recent survey year [defaulted for prevalence], default select that year for Time 1 (both prevalence and ART coverage). If no data in that year, do not select default
* If ANC data exist in the year 2020, preselect that for 'Time 2' in ANC clients, ANC prevalence, and ANC coverage.

Also taken opportunity here to tidy up the model fit completed text as it was wrapping over 2 lines on macs which didn't look amazing


jeffeaton (approved):
Merci!

Issue #180, opened 2020-12-04
catch NA quantile error when simulating outputs and return useful message


jeffeaton:
This error occurs when the model simulates and NA value for an indicator in producing the outputs package.
```
converged: false convergence (8)
Error in quantile.default(newX[, i], ...) : 
  missing values and NaN's not allowed if 'na.rm' is FALSE
```

This error should be caught and return which indicator threw the error.

PR #196, opened 2020-12-12
handle case when 100% Spectrum ART coverage -> no untreated PLHIV denominator

jeffeaton:
This PR:

* Catch error when `quantile()` encounters `NA` during `output_package()` creation and return an error message indicating which indicator threw the error. 
  * Error message is tested. 
   * Addresses #180. 

* Converts `$` to `[[` in several places to avoid auto completion errors.

* If ART coverage is 100% in an age/sex category, resulting in zero untreated PLHIV denominator, replace the `NaN` `unaware_untreated_prop` with 1.0. This addresses support ticket 11 from Zambia team.


r-ash (approved):
This looks good, we could turn on warnings for partial matching if helpful? Might just generate a load of noise form dependencies though..

PR #199, opened 2020-12-18
MRC 2078: Add calibration options back in

r-ash:
This PR will
* Revert #191
* Also throws an error if any required calibration options are missing

We should leave this unmerged until HINT devs tell us all is good to go.


r-ash:
This is ready for review now we have deployed hintr & hint to staging and seen this working

richfitz (approved):

PR #213, opened 2021-01-22
Make hintr_calibrate write to new files and not overwrite existing

r-ash:


richfitz (approved):
LGTM. This will presumably make things take more space, but we use so little we can deal with that down the track. It will be interesting to look later to see how many are identical

PR #223, opened 2021-02-09
Calibration options fix

rtesra:


R/run-model.R
```diff
@@ -564,22 +564,22 @@ format_options <- function(options) {
   if (is.null(options$spectrum_artnum_calibration_level)) {
     options$spectrum_artnum_calibration_level <- "none"
   }
-  if (is.null(options$spectrum_artnum_strat)) {
+  if (is.null(options$spectrum_artnum_calibration_strat)) {
```
jeffeaton, line 567: These fixes should be tested. (Lack of testing of these functions is why the error happened in the first place...).

Let's get this fix merged, so approving without the tests. But can you create an issue to add tests for this function?

inst/report/summary_report.Rmd
```diff
@@ -380,8 +380,19 @@ cat(paste0("\n ",cat(paste0("* ", text2$prefix, "_", text2$source, "_"), sep = "
...
+# Read in calibration options from calibration_options.yml, if that does not exist, 
+# use default options in options.yml
+
+calibration_files <- utils::unzip(output_zip, list = TRUE)
+
+if("info/calibration_options.yml" %in% calibration_files$Name) {
+  calibration_options <- unz(output_zip, "info/calibration_options.yml")
+  calibration_options <- yaml::read_yaml(calibration_options)
+} else {
+  calibration_options <- options[grep("calibration", names(options))]
```
jeffeaton, line 392: This code will not include the `calibrate_method` option. I would like to add that back into the report for completeness. Can you revise to add that back in?

jeffeaton (changes requested):
Thanks, these changes look good.

* The fix to `format_options()` should be tested. (Lack of testing this function is why the error was able to creep through... :/). I'm happy to merge this fix without adding test, but can you create an issue to add tests for this?

* I would like to add the `calibrate_method` option back into the summary report, which removed last Friday to fix Rob G.'s issue.  Can we pull that into this update? Rob's error was because there was a `NULL` for `options$calibrate_method`.  The logic that needs adding is if `calibrate_method` is not in the options.yml, a value needs to be set for it.

R/run-model.R
rtesra: Issue added here: #224

PR #230, opened 2021-03-19
Name all translator_register args

r-ash:
Thanks @krisher1 for reporting this

We changed the options in `traduire` to hopefully be more extensible in the future. The result is that the args to `translator_register` need to be named.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -8,7 +8,9 @@ naomi_init_traduire <- function() {
   pattern <- sprintf("%s/{language}-{namespace}.json", root)
   languages <- c("en", "fr")
   namespaces <- "translation"
-  traduire::translator_register(NULL, languages[[1]], namespaces[[1]],
+  traduire::translator_register(resources = NULL,
+                                language = languages[[1]],
+                                default_namespace = namespaces[[1]],
                                 resource_pattern = pattern,
                                 namespaces = namespaces,
                                 fallback = "en",


jeffeaton:
Looks good.  I can't tell what the error was that identified this though. Is that important?

r-ash:
> Looks good. I can't tell what the error was that identified this though. Is that important?

Thanks, useful to have it here for history sake

```
installing to /Library/Frameworks/R.framework/Versions/4.0/Resources/library/00LOCK-naomi-2.3.1/00new/naomi/libs
** R
** data
*** moving datasets to lazyload DB
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
Error: package or namespace load failed for ‘naomi’:
 .onLoad failed in loadNamespace() for 'naomi', details:
  call: NULL
  error: 'options' must have unique names
Error: loading failed
Execution halted
ERROR: loading failed
* removing ‘/Library/Frameworks/R.framework/Versions/4.0/Resources/library/naomi’
Warning message:
In install.packages("~/Downloads/naomi-2.3.1.tar.gz") :
  installation of package ‘/Users/kathrynrisher/Downloads/naomi-2.3.1.tar.gz’ had non-zero exit status
```

Would have occured on package load too - key part is `error: 'options' must have unique names`. `traduire` will deliberately fail if args aren't named now

jeffeaton:
Thanks for adding the information here.

jeffeaton (approved):

PR #241, opened 2021-05-11
Add function to return data for plotting calibrate barchart

r-ash:
Need to tidy up the spectrum data we are returning here still.

Also there might be a better way to generate this

This is on hold for the moment, waiting on some additional data to be added to `spectrum_calibration` table in output package


r-ash:
Superseded by #261

PR #243, opened 2021-05-17
mrc-2406: Support on demand generation of model outputs

r-ash:
This PR will
* Add functions `hintr_prepare_spectrum_download`, `hintr_prepare_coarse_age_group_download` and `hintr_prepare_summary_report_download` for generating downloads from `hintr_output` object
*  Refactor `hintr_run_model` and `hintr_calibrate` to save out minimal data (they now save out only what was the `calibration_data` and `output_path`

We should not merge this until front end has been updated to support https://github.com/mrc-ide/hintr/pull/279

This is a fairly big change to the contract, it will break all previous model results. We have a migration script in `hintr` PR which will manage updating so previous results aren't broken. Note that the migration will regenerate their outputs from the `calibration_data`so expect the report to change for really old model results.

I think this will break some of the orderly tasks, if you need something to sit in the equivalent of `hintr_run_model` was you can use
```
model_output <- hintr_run_model(data, options)
calibrated <- hintr_calibrate(model_output, calibration_options)
spectrum <- hintr_prepare_spectrum_download(calibrated)
coarse_age <- hintr_prepare_coarse_age_group_download(calibrated)
summary_report <- hintr_prepare_summary_report_download(calibrated)
list(
  output_path = calibrated$plot_data_path
  spectrum_path = spectrum$path,
  coarse_output_path = coarse_age$path,
  summary_report_path = summary$path,
  calibration_path = calibrated$model_output_path
)
```
Or we could add a helper for this


richfitz (approved):

r-ash:
> Thanks. Corresponding hintr PR at [mrc-ide/hintr#278](https://github.com/mrc-ide/hintr/pull/278)

Closed this PR in favour of https://github.com/mrc-ide/hintr/pull/283

PR #244, opened 2021-05-21
Install info

athowes:
* Added information about how to install the package to README
* Some small formatting things on README
* Ran make website (which updated one of the vignettes, not that I changed a vignette) as I thought this would be required for the README change to take effect but not so sure that's how it works now


README.md
```diff
@@ -5,16 +5,26 @@
...
+
 ### Development steps
 
 * Make changes in a new branch
 * Run tests from the command line using `make test` or via R using `devtools::test()` to ensure changes don't cause regressions to existing functionality
 * Run build and check before pushing to remote using `make build` and `make check` from command line or `devtools::check()` from R
 * When branch is ready for merging create a PR and add a reviewer
 * Ensure that the version number has been updated according to [semantic versioning](https://semver.org/) and add a news item describing the change
-* Reviewer should check code and ensure the build passes on [travis](https://travis-ci.org/mrc-ide/naomi) before merging
+* Reviewer should check code and ensure the build passes on [Travis](https://travis-ci.org/mrc-ide/naomi) before merging
```
r-ash, line 25:
```suggestion
* Reviewer should check code and ensure the build passes on [Buildkite](https://buildkite.com/mrc-ide/naomi) before merging
```

r-ash: We're actually using buildkite for this now, would be good to update it whilst we're here

r-ash (approved):
Thanks for this, looks like vignette output was just errors before so good to have that fixed too

PR #249, opened 2021-07-30
manually avoid write_datapack_csv() warning when fitting with zone de…

jeffeaton:
See branch name for description.

Quite a manually solution. A better solution is to change the ISO3 for the demo data to an artificial code. However, some hint validation needs to be relaxed first to enable this.


r-ash (approved):
:pray:

PR #250, opened 2021-08-03
Add functions for shaping input time series data

r-ash:
This PR will add function `prepare_input_time_series_anc` and `prepare_input_time_series_art` which take art/anc data and shape file and output format needed for plotting input time series (shinyRob)


R/input-time-series.R
```diff
@@ -0,0 +1,170 @@
...
+      dplyr::rename(time_period = calendar_quarter) %>%
+      rbind(art_number_annual)
+  } else if (unique(dup$n == 1)) {
+    # If annual values available for all disags - don't aggregate
+    art_number <- art_number %>%
+      dplyr::mutate(time_step = "annual") %>%
+      dplyr::rename(time_period = calendar_quarter)
+  } else {
+    # Throw error for duplicate annual or incomplete quarterly data
+      stop(print("Quarterly data not provdied for all dissagregates/n
```
rtesra, line 53: @jeffeaton @r-ash Should we add a check in here that would throw an error and identify:
1. Missing quarterly data
2. Duplicate annual data
Or should user be able to upload incomplete data at this stage?

jeffeaton:
I haven't reviewed the code yet, but quickly my opinions on these questions:


  1.  Missing data should be allowed, but should go to an NA value in an aggregate, e.g. sum(x, na.rm = FALSE)​.  If users think missing data should be 0 instead of missing, then enforce a 0 to be coded.
     *   Differential missing data in numerator / denominator might need some thinking through.
  2.  Explicit duplicate data should not be allowed -- should throw an error. E.g. if two values are entered for males 15+ in district X --> error.
     *   Similarly, I think that duplicated data at different levels should through an error, e.g. data entered for district X and for province Y contained in district X.  This could be debated. There is already some code written to check this sort of duplication.

Thanks,
Jeff
________________________________

R/input-time-series.R
r-ash: So any data that gets here will have already had this function https://github.com/mrc-ide/hintr/blob/master/R/validate_inputs.R#L112 called on it. Or for anc https://github.com/mrc-ide/hintr/blob/master/R/validate_inputs.R#L149

So for ART
1. Not checking for missing quarterly data - we could add this as a check in the validate function
2. We're checking that there is a unique combination of `"area_id", "calendar_quarter", "sex", "age_group"` (not sure if that covers duplicate annual case?) 

For ANC we check that there are unique combinations of `"area_id", "age_group", "year"`.

If there are any data validation checks which we need to run then we should put them into the validate functions - so that users will get the error on upload instead of when trying to generate the graph.

r-ash: We should also use translated text for the error `stop(t_("KEY"))`

R/input-time-series.R
```diff
@@ -0,0 +1,186 @@
...
+##' area_id, area_name, area_level_label, time_step, time_period, plot and
+##' value
+##' @export
+prepare_input_time_series_art <- function(art, shape) {
+
+  ## From "area_id", "calendar_quarter", "sex", "age_group", "art_current"
+
+  ## To "area_id","area_name","area_level_label", "area_level", time_step",
+  ## "time_period","plot", "value"
+
```
r-ash, line 19:
```suggestion
```

r-ash: These were just notes I made when there was no impl here but I think not useful anymore.

R/input-time-series.R
```diff
@@ -0,0 +1,186 @@
...
+  ## Check if data is quarterly or annual - if quarterly aggregate to annual
+  areas <- sf::read_sf(shape) %>% sf::st_drop_geometry()
+
+  art_number <- readr::read_csv(art, show_col_types = FALSE) %>%
```
r-ash, line 23: Should we use `read_art_number` for reading this data?

R/input-time-series.R
```diff
@@ -0,0 +1,186 @@
...
+
+  ## TO "area_id","area_name","area_level","area_level_label","age_group",
+  ## "time_period","time_step","plot","value"
+
+
+  ## Recursively aggregate ANC data up from lowest level of programm data provided
+  # Level to aggregate from
+  areas <- sf::read_sf(shape) %>% sf::st_drop_geometry()
+
+  anc_testing <- readr::read_csv(anc, show_col_types = FALSE) %>%
```
r-ash, line 139: Similarly should we use `read_anc_testing` here?

R/input-time-series.R
```diff
@@ -0,0 +1,186 @@
...
+      dplyr::rename(time_period = calendar_quarter) %>%
+      rbind(art_number_annual)
+  } else if (unique(dup$n == 1)) {
+    # If annual values available for all disags - don't aggregate
+    art_number <- art_number %>%
+      dplyr::mutate(time_step = "annual") %>%
+      dplyr::rename(time_period = calendar_quarter)
+  } else {
+    # Throw error for duplicate annual or incomplete quarterly data
+      stop(print("Quarterly data not provdied for all dissagregates/n
```
r-ash, line 53: Can we put this through traduire? Using `stop(t_("KEY"))`

R/input-time-series.R
```diff
@@ -0,0 +1,186 @@
...
+##' time_step, plot and value
+##' @export
+prepare_input_time_series_anc <- function(anc, shape) {
+  ## From "area_id", "age_group", "year", "anc_clients",
+  ## "anc_known_pos", "anc_already_art", "anc_tested", "anc_tested_pos"
+
+  ## TO "area_id","area_name","area_level","area_level_label","age_group",
+  ## "time_period","time_step","plot","value"
+
+
```
r-ash, line 134:
```suggestion
```

tests/testthat/test-input-time-series.R
```diff
@@ -0,0 +1,121 @@
...
+  ## determined by highest area_level in program data
+  shape <- sf::read_sf(a_hintr_data$shape)
+  art_number <- art_number %>%
+    dplyr::left_join(shape %>% dplyr::select(area_id, area_level), by = "area_id")
+
+  # Check data has been aggregated from correct baseline
+  expect_equal(unique(art_number$area_level), max(data$area_level))
+  # Check data has been aggregated from baseline to lowest level in hierarchy
+  shape_level <- unique(shape$area_level)[unique(shape$area_level) <= unique(art_number$area_level)]
+  expect_equal(shape_level, unique(data$area_level))
```
r-ash, line 55: Sorry this is a bit picky but can you switch these around?
```suggestion
  expect_equal(unique(data$area_level), shape_level)
```
convention is to have the thing you are testing on the LHS.

r-ash (commented):
Thanks Rachel, this looks good - tests are great.

Couple of small comments.

R/input-time-series.R
r-ash: No - this doesn't work because it drops some columns we need

R/input-time-series.R
r-ash: This also doesn't work because it drops some columns we need

richfitz (approved):

PR #253, opened 2021-09-08
Add translated human readable names for plot type

r-ash:
PR will
* Return translated human readable names for plot type from input time series functions

@rtesra Can you check I have created appropriate labels (and suggest better ones where they aren't very good)? And I have added a translated string for each of the plot types (note a couple already existed in the list of strings)


r-ash:
Closing this in favour of #254

PR #254, opened 2021-09-20
mrc-2614: Add function to get translated label and desc for plot types

r-ash:
This replaces #253 sorry for any confusion

Want to get review from science on the labels and description text. Please feel free to edit these files and push edits to this PR.


tests/testthat/test-input-time-series.R
```diff
@@ -152,3 +152,19 @@ test_that("ART data throws error if dupe annual data or incomplete quarterly", {
...
+  expect_equal(ret, list(
+    list(
+      id = "art_total",
+      label = "ART count",
+      description = "Number on ART at the end of calendar year"
+    ),
+    list(
+      id = "art_child",
+      label = "ART paediatric",
+      description = "Proportion of total on-ART under age 15"
```
mwoodbri, line 167:
```suggestion
      description = "Number of children (<15) on ART at the end of calendar year"
```
Not clear whether this should be "Proportion" or "Number" but test failure is due to mismatch with JSON

inst/traduire/en-translation.json
```diff
@@ -233,6 +233,28 @@
...
+    "ART_TOTAL": "ART count",
+    "ART_TOTAL_DESC": "Number on ART at the end of calendar year",
+    "ART_ADULT_F": "ART female",
+    "ART_ADULT_F_DESC": "Number of women (15+) on ART at the end of calendar year",
+    "ART_ADULT_M": "ART male",
+    "ART_ADULT_M_DESC": "Number of men (15+) on ART at the end of calendar year",
+    "ART_ADULT": "ART adult",
+    "ART_ADULT_DESC": "Number of adults (15+) on ART at the end of calendar year",
+    "ART_CHILD": "ART paediatric",
+    "ART_CHILD_DESC": "Number of children (<15) on ART at the end of calendar year",
```
richfitz, line 246: Looks like we've got many more keys added here than in the fr and pt translations - what is our preferred workflow here? To add dummy keys, google translations, or leave as-is? I think that as-is the key will get reflected out into the application, which is fine for now. We could also perhaps more easily detect missing keys by looking to see what is present in the en set but missing from the others?

NAMESPACE
```diff
@@ -8,6 +8,7 @@ export(anc_testing_prev_mf)
 export(artnum_mf)
 export(calculate_prevalence_art_coverage)
 export(calendar_quarter_labels)
+export(calendar_quarter_labels_short)
```
richfitz, line 11: Why is this exported here? unrelated change or needed in a dependent hintr pr?

richfitz (commented):
two small queries here, code itself looks fine

inst/traduire/en-translation.json
r-ash: Plan is at some point before estimates round (probably some time early Nov) do a diff from the last date we received translations to now. Then add those new keys to a spreadsheet for UNAIDS to translate.

Have previously to this point been adding translations from google or deepl so it is at least partially usable by people but after discussion the other day maybe this is not worth the effort as it is quite boring to add these placeholder translations.

NAMESPACE
r-ash: Unrelated - the `@export` tag was added to docs at some point but they must not have regenerated

richfitz: ah, ok

PR #255, opened 2021-09-28
Include parent area id in response from time series

r-ash:
This is so that in hintr we can construct the hierarchy for returning to the front end

This is better than doing it in hintr separately as it means we can avoid reading the shape file twice and means the hierarchy is already filtered to the level in the available ART/ANC data


rtesra:
Looks good

rtesra (approved):

PR #256, opened 2021-10-19
mrc-2663: Return naomi warnings in the output from model fit, model calibrate and model option validation

r-ash:
This PR will
* Add `warnings` item to list returned from `hintr_run_model`, `hintr_calibrate` and `validate_model_options`
* Adds an example of how to use `naomi_warning` for science
 
For @rtesra or @jeffeaton 
* We need to add some real translated warnings, either to this PR or to a PR which builds on this. We also need some data or set of options we can choose which will trigger at least one of the warnings from each of the functions which we can use for testing and verifying after we deploy
* Do you want these warnings to be printed to someone calling the function? At the moment they won't be


richfitz (approved):
This looks great to me, leaving open in case science have things to add

.Rprofile
```diff
@@ -0,0 +1,3 @@
+options(
+  warnPartialMatchArgs = FALSE
```
richfitz, line 2: this is possibly problematic as I think it prevents reading of the user Rprofile - what is this trying to achive?

richfitz (changes requested):

.Rprofile
r-ash: Oops, I pushed this by mistake. Thanks

R/run-model.R
```diff
@@ -76,8 +89,8 @@ hintr_run_model <- function(data, options,
                  max_iter = options$max_iterations %||% 250,
                  progress = progress)
 
-  if(fit$convergence != 0 && !options$permissive) {
-    stop(paste("convergence error:", fit$message))
+  if(fit$convergence != 0) {
+    naomi_warning(paste("convergence error:", fit$message), "model_fit")
   }
```
rtesra, line 94: @jeffeaton do we want to totally remove the error thrown by false convergence (implemented here) or still have the option with permissive = false?

jeffeaton: I am happy to remove the error and fully convert it to a warning

r-ash: Should we also remove `permissive` option?

jeffeaton: I just did a quick search through the code -- the `permissive` option is not used anywhere else, so yes fine to remove. It's also not a big deal to keep it; it is pretty hidden away from user.

richfitz (approved):

PR #267, opened 2021-11-17
Retrire viridis

r-ash:
:see_no_evil:


r-ash:
After discussion with UNAIDS they have some issues with the proposed new colours, they think it is hard to tell differences when in the white end of the spectrum. They want to do something like only use the darker 75% of the colour scale but not something we can get in before countries use the app. Marking this as blocked until we get some traction from UNAIDS or some further discussion

rtesra (approved):

PR #269, opened 2021-11-18
Remove unnecessary text in tooltip

r-ash:
Feedback from user session, the first sentence in the tooltip is not needed.
Screenshot_20211118_144137

Another comment was on the number of survey options we show (option "Calendar quarter at midpoint of survey"), this is controlled from hintr but it hardcoded to show all quarters from 2010 to now. Should we reduce the number shown? Same for the "Calendar quarter to generate estimates"?


jeffeaton:
Thanks the changed text looks good.

> Another comment was on the number of survey options we show (option "Calendar quarter at midpoint of survey"), this is controlled from hintr but it hardcoded to show all quarters from 2010 to now. Should we reduce the number shown?

This cannot be fewer -- I think the earliest surveys we have are back in 2012 or so.

> Same for the "Calendar quarter to generate estimates"?

This could probably start from 2018 or so. I don't know that it is a big deal though. It defaults to `CY2021Q4`, and very large majority of users will not change that.

jeffeaton (approved):
Looks good

PR #273, opened 2021-11-26
Update metadata for input time series to include format and accuracy

r-ash:
Don't merge this until Emma has confirmed it meets what they need


r-ash:
Here can @jeffeaton or @rtesra review the metadata for the time series indicators and set appropriate format & accuracy. This conforms to same format as other metadata. Format is a http://numeraljs.com/ format, and accuracy rounds to specified level. Blank means no rounding and e.g. accuracy of 100 would display 342234 as 342200 for example.

rtesra (approved):

PR #274, opened 2021-11-28
Naomi v2.6.0 update

jeffeaton:
This PR makes several small model updates for the 2022 estimates round:

* In construction of projection matrices, single-year populations are interpolated to quarter-year using monotonic splines instead of constant allocation within each age to slightly smooth out the ageing process.
* ~~A term is added to account for district-level net migration by cohort in the PLHIV projection. This reduces cases where there are sharp changes in prevalence due to change in the district population (but previously not PLHIV).~~ ***This update is pushed to next version.***
* Spectrum Dec 31 ART inputs are used in the ART calibration, so that Q4 Naomi results calibrated to Spectrum will exactly match Spectrum ART inputs.
* Paediatric new infections (MTCT) are incorporated as the ratio of paediatric HIV incidence vs. adult female (15-49) prevalence.

An internal change is that the R version of the Naomi model has been added as a function `naomi_objective_function_r()`.  
* This is primarily for stepping through the model line-by-line for debugging and model development. 
* There is a test to ensure the `REPORT()` outputs of the R version match the C++ version.
* The value of the objective function is not tested yet because I still need to implement TMB's AR1 model in R. There is a test for this which is skipped.


r-ash:
hintr build passing so I think good to merge when @rtesra is happy

jeffeaton:
I'm adding _blocked_ label on this. Rachel's test runs identified this results in `NaN` likelihood evaluation for several countries.

I've debugged it for BDI. The issue there is related to the change to use number on ART in Q4 from Spectrum input. The number is larger than number of PLHIV, resulting in 0 untreated population for women of childbearing age --> NaN for ART fertility rate ratio calculation from Spectrum.

I will try to fix this later today to get an updated version to run through the tests again.

jeffeaton:
Hi @r-ash, @rtesra,

The cohort growth ratio adjustment needs a bit more thought, but we need to get the other updates in here deployed. I have removed the effect of that from this v2.6.0 update, and will aim to put it in next PR.  

A couple of the tests that check the parameter `adjust_area_growth = TRUE` affects the projection matrices are skipped.

Thanks,
Jeff

r-ash (approved):

PR #276, opened 2021-11-29
suppress s2 geometry in adjacency matrix construction

jeffeaton:
What do you guys think about adding this bit of code to suppress using `s2` geometry in the adjacent matrix construction?   This addresses fitting issue with flamingo.

It is currently a bit chatty:
```
> naomi_data <- naomi:::naomi_prepare_data(data, options)
Spherical geometry (s2) switched off
although coordinates are longitude/latitude, st_intersects assumes that they are planar
Spherical geometry (s2) switched on
```

We could catch and suppress those messages if they are a nuisance.


r-ash (approved):
Looks good to me, if chattiness gets annoying we can quiet it down later

jeffeaton:
> Looks good to me, if chattiness gets annoying we can quiet it down later. Could you fix conflicts?

Update: they got annoying. I just updated to suppress them.

For some reason the message from `spdep::poly2nb()` wasn't being quieted using the `withCallingHandlers()` call to silence a single message; I wrapped it in `suppressMessages()` which I think should be benign for this case (just messages, not warnings being quieted).

I also merged master.

Could you please re-review and merge if satisfied?

PR #277, opened 2021-12-01
Filter on max art_level to avoid possibility of using vector in filter value

r-ash:
This will fix issue 38, caused by them uploading ART with more than 1 level in it. I will add a validation check to hintr too to stop people from doing this earlier on


jeffeaton (approved):
If I were being pedantic, I would ask to see the regression test!  (see how well you have trained me.)

rtesra (approved):

PR #278, opened 2021-12-01
first90 v1.5.1

jeffeaton:
This PR updates the first90 package dependency to v1.5.1.

This is incremental on #271, but I created it as a separate PR because I can't remember what needs to be done in Buildkite and front end to update first90 package, so thought it better to keep it separate.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,7 @@ Imports:
     data.tree,
     dplyr,
     eppasm (>= 0.5.9),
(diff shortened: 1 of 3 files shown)


jeffeaton:
As expected, this failed due to not current package version on build kite:
```
* checking package dependencies ... ERROR
--
  | Package required and available but unsuitable version: ‘first90’
```

r-ash (approved):

PR #279, opened 2021-12-06
Navigator checklist

rtesra:


R/run-model.R
```diff
@@ -600,7 +600,8 @@ naomi_info_packages <- function() {
 naomi_info <- function(data, options) {
   list("inputs.csv" = write_csv_string(naomi_info_input(data)),
        "options.yml" = yaml::as.yaml(options),
-       "packages.csv" = write_csv_string(naomi_info_packages()))
```
rtesra, line 653: @jeffeaton data added into info here

R/outputs.R
```diff
@@ -763,13 +784,22 @@ save_output <- function(filename, dir,
...
-    for (p in names(info)) {
-      writeLines(trimws(info[[p]]), file.path("info", p))
+
+  options <- yaml::read_yaml(text = info$options.yml)
+  data <- info$data
+
+  write_navigator_checklist(naomi_output, options, data,
+                            "info/unaids_navigator_checklist.csv")
+
+  info_sub <- info
```
rtesra, line 798: @jeffeaton here is where info is saved back out - added this in to stop the additional `data` element from being saved out

r-ash (changes requested):
Some comments, some are about formatting which is quite pedantic but I think can go a long way to making code easier to read.

This is also quite light on testing, there is quite a lot of business logic in the `write_navigator_checklist` which would be good to check we have right.

Feel free to ignore some of the comments, I recognise we are tight for time, but I think we should tidy up the format of the code and add some tests. Definitely of the backward compatibility as atm I think this will return a hard to understand error to the user.

R/run-model.R
jeffeaton: This is removed by refactored code.

R/outputs.R
jeffeaton: `data` no longer saved in `info` object in re-factored code.

R/unaids-navigator.R
```diff
@@ -53,16 +58,111 @@ write_navigator_checklist <- function(naomi_output,
...
                   NaomiCheckDes = label,
-                  TrueFalse = NA)
+                  TrueFalse = FALSE)
 
+  ## Do checks
   ## These checks are always 'TRUE'
   always_true_checks <- c("Package_created", "Package_has_all_data", "Model_fit")
   v$TrueFalse[v$NaomiCheckPermPrimKey %in% always_true_checks] <- TRUE
 
+  # TODO: Compare aggregated naomi inputs to national Spectrum totals
```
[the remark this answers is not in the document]
jeffeaton: Thanks -- yes. I'm working on this one.

R/outputs.R
```diff
@@ -843,9 +874,12 @@ read_output_package <- function(path) {
   calibration_options <- readr_read_csv(file.path(tmpd, "fit/calibration_options.csv"))
   calibration_options <- setNames(calibration_options$value,
                                   calibration_options$option)
+  model_options <- yaml::read_yaml(file.path(tmpd,"info/options.yml"))
```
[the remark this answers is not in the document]
jeffeaton: code removed--resolving

R/unaids-navigator.R
```diff
@@ -1,13 +1,18 @@
 #' Write UNAIDS Estimates Navigator checklist CSV
 #'
 #' @param naomi_output a naomi_output object.
-#' @param path path to save Navigator checlist CSV.
+#' @param options naomi model options
```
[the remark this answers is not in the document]
jeffeaton: Code removed. `calibration_options` is now taken from the `naomi_output` object

R/unaids-navigator.R
```diff
@@ -53,16 +58,111 @@ write_navigator_checklist <- function(naomi_output,
...
+  # Is most recent calendar quarter selected
+  if(options$calendar_quarter_t2 == valid_opt$calendar_quarter_t2){
+    v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_recent_qtr"] <- TRUE}
+
+  # Is future projection quarter selected
+  if(options$calendar_quarter_t2 == valid_opt$calendar_quarter_t2){
+    v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_future_proj_qtr"] <- TRUE}
+
+  # Is ART attendance selected
+  if(options$artattend){v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_ART_attendance_yes"] <- TRUE}
```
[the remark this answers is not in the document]
jeffeaton: A bit of an aside, but I prefer this one:
```
v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_ART_attendance_yes"] <- options$artattend
```
over this one:
```
v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_ART_attendance_yes"] <- options$artattend
```

Curious why your preference for the latter?  (The latter is a bit safer, because it does not risk auto-completion of list elements; in particular it is much safer to write `options$artattend` as `options[["artattend"]]`, since there is another list element `options$artattend_t2`.  I remember exactly where I was sitting--on my sister's sofa--when I debugged that over Christmas 2019.)

R/unaids-navigator.R
```diff
@@ -53,16 +58,111 @@ write_navigator_checklist <- function(naomi_output,
...
+
+  # Is scope set to national ID
+  scope <- options$area_scope
+  national_id <- naomi_output$meta_area$area_id[naomi_output$meta_area$area_level == 0]
+  if(scope == national_id){v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_area_ID_selected"] <- TRUE}
```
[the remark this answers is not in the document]
jeffeaton: It has been updated to the first throughout

R/unaids-navigator.R
```diff
@@ -53,16 +58,111 @@ write_navigator_checklist <- function(naomi_output,
...
+
+  ## Check survey
+  # Does T1 calendar quarter match one survey
+  t1 <- gsub(".*?([0-9]+).*", "\\1", options$calendar_quarter_t1)
```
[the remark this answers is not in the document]
jeffeaton: There is an existing function `year_label()` for this [could be better named...]. But comment obsolete as we should check the exact quarters here, not string-parse for years.

R/unaids-navigator.R
```diff
@@ -53,16 +58,111 @@ write_navigator_checklist <- function(naomi_output,
...
+  surveys <- gsub(".*?([0-9]+).*", "\\1", options$survey_prevalence)
+  if(t1 %in% surveys){v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_calendar_survey_match"] <- TRUE}
+
+  # Is most recent survey in input data selected in options
+  input_survey <- read.csv(data$survey$path)
```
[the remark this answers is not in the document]
jeffeaton: Obsolete--no longer read in these data; information comes from new `naomi_output$fit$data_options` list.

R/unaids-navigator.R
```diff
@@ -53,16 +58,111 @@ write_navigator_checklist <- function(naomi_output,
...
+  # Is survey ART coverage included if available
+  input_survey_ind <- unique(input_survey$indicator)
+  if("art_coverage" %in% input_survey_ind && !is.null(options$survey_art_coverage)) {
+    v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_ART_coverage"] <- TRUE}
+
+  ## Check ART inputs
+  # Is ART data included if available
+  art_options <- c(options$include_art_t1, options$include_art_t2)
+
+  if(TRUE || "true" %in% art_options && !is.null(data$art_number)) {
```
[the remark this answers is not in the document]
jeffeaton: :-/

jeffeaton:
Hi @r-ash, @rtesra,

This is ready for review now. I have added code to extract the Spectrum inputs for ANC testing and number on ART (in `inputs-spectrum.R`), and added code in `select_naomi_data()` to check if the aggregated Naomi inputs match the Spectrum inputs. For backward compatibility and not obstructing other code, the function still runs `select_naomi_data()` still runs if the `spec_program_data` object is not provided (defaults to `NULL`) and the function returns `NA` for the equality checks in the `data_options` return.

It is slightly inefficient because the PJNZ gets read from twice to extract the inputs required for Naomi and then in a second function to extract these programme data inputs. But I think it is a right balance between streamlined implementation and not having knock on effects for other parts of the code.

The tests are not perfect—they do not check the TRUE/FALSE logic for every condition. But they do confirm the condition gets checked (i.e. does not return an `NA` or `NULL` for any checks) with a standard model fit and constructs a set of options in the output object that returns `TRUE` for all conditions.

I think very few countries will pass the check for Naomi and Spectrum inputs matching exactly for all years. I think UNAIDS et al. may want to relax that a bit; but let's implement as specified for now and can discuss with them later about relaxing.

Thanks,
Jeff

R/outputs.R
```diff
@@ -763,25 +783,27 @@ save_output <- function(filename, dir,
...
+  dir.create("info")
   info <- attr(naomi_output, "info")
+
   if (length(info) > 0L) {
     for (p in names(info)) {
       writeLines(trimws(info[[p]]), file.path("info", p))
     }
   }
 
+  write_navigator_checklist(naomi_output,
```
r-ash, line 796: This will fail for old model outputs I think (because data options are not available) - probably with a not very useful error. We should check if the `naomi_output` has the appropriate format to create the checklist. If not then don't save this file out, and maybe display a warning to the user?

r-ash: We need a test that checks this case too

R/unaids-navigator.R
```diff
@@ -53,16 +54,147 @@ write_navigator_checklist <- function(naomi_output,
...
+  calibration_options <- naomi_output$fit$calibration_options
+
+  ## Compare aggregated naomi inputs to national Spectrum totals
+
+  v$TrueFalse[v$NaomiCheckPermPrimKey == "ART_is_Spectrum"] <- data_options$art_number_spectrum_aligned
+  v$TrueFalse[v$NaomiCheckPermPrimKey == "ANC_is_Spectrum"] <- data_options$anc_testing_spectrum_aligned
+
+
+  ## Is most recent calendar quarter selected
+  if (naomi_output$fit$model_options$calendar_quarter_t2 == valid_opt$calendar_quarter_t2) {
```
r-ash, line 83: I think these are better as `v$TrueFalse[v$NaomiCheckPermPrimKey == "Opt_recent_qtr"] <- naomi_output$fit$model_options$calendar_quarter_t2 == valid_opt$calendar_quarter_t2` (and that means code cov won't flag these lines as missing)

R/unaids-navigator.R
```diff
@@ -53,16 +54,147 @@ write_navigator_checklist <- function(naomi_output,
...
+  # Is logistic calibration selected
+
+  calibration_options <- naomi_output$fit$calibration_options
+
+  if (calibration_options$calibrate_method == "logistic") {
+    v$TrueFalse[v$NaomiCheckPermPrimKey == "Cal_method"] <- TRUE
+  }
+
+
+  if (all(naomi_output$meta_area$spectrum_region_code == 0)) {
```
r-ash, line 155: This is the kind of forking which we want a test to cover. Ideally every path should be run by the test code. So we want a test where `spec_level <- "nat"` and one where `spec_level <- "subnat"` so that both paths of the if on line 161 get run. Otherwise we are at risk of a silly typo in one of the names or something breaking the output

R/inputs-spectrum.R
```diff
@@ -95,6 +95,252 @@ extract_pjnz_one <- function(pjnz) {
...
+  art_dec31 <- rbind(child_art, art15plus)
+
+  art_dec31
+}
+
+#' Disaggregate the number on ART Dec 31 to single age
+#'
+#' @noRd
+#' 
+add_dec31_art <- function(spec, pjnz) {
```
r-ash, line 256: Maybe me being silly but I can't see where this function is used?

r-ash (changes requested):

R/unaids-navigator.R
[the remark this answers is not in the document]
jeffeaton: This is completed now -- closing

R/inputs-spectrum.R
jeffeaton: Sorry -- this code got copied over from the v2.6.0 update in #274; it is used there. I think it will create a bit of merge havoc to remove it here now (which will later get merged into that branch). Okay to leave it in here?

R/unaids-navigator.R
jeffeaton: addressed by 4143160

R/outputs.R
jeffeaton: Thanks -- this is addressed in 10e2441.  The behaviour is to still write the file, but `NA` values are returned for checks that cannot be checked due to missing options list. 

Tests are here: https://github.com/mrc-ide/naomi/blob/navigator-checklist/tests/testthat/test-outputs.R#L320-L387

R/unaids-navigator.R
r-ash: Logic is fairly simple both sides so should be fine. Let's get this in

r-ash (approved):

PR #283, opened 2021-12-08
Use standard file reading functions in input time series

r-ash:


R/input-time-series.R
```diff
@@ -320,7 +320,7 @@ prepare_input_time_series_anc <- function(anc, shape) {
 
   ## Check if art is object or file path
   if(!inherits(anc, c("spec_tbl_df","tbl_df","tbl","data.frame" ))) {
-    anc <- readr::read_csv(anc, show_col_types = FALSE)
+    anc <- read_anc_testing(anc)
```
jeffeaton, line 323: Do you need the `area_name` from this one? (Does that get used in any of the time series code, etc.?)

rtesra (approved):

jeffeaton (approved):
Thanks, nice one.  I am approving this based on what is here looks good.

One thing to check: does the time series plots rely on the `area_name` at all?  
* Will not having the `area_name` field read by `read_anc_testing()` create any issue?
* Will having the `area_name` present in `read_art_number(..., all_columns = TRUE)` create any problem? (E.g. merging with another dataset that has `area_name`)

R/input-time-series.R
rtesra: No `area_name` is pulled back in after merging on `area_id` with the boundaries file so not needed from the anc data:  https://github.com/mrc-ide/naomi/pull/283/files#diff-41ca6a0b90b4c5a022737ba75b3a161ccd69eaa2258feb8cb628ed2bd69733bdR30

r-ash:
> Thanks, nice one. I am approving this based on what is here looks good.
> 
> One thing to check: does the time series plots rely on the `area_name` at all?
> 
>     * Will not having the `area_name` field read by `read_anc_testing()` create any issue?
> 
>     * Will having the `area_name` present in `read_art_number(..., all_columns = TRUE)` create any problem? (E.g. merging with another dataset that has `area_name`)

Thanks, have updated to remove `area_name` reading from `read_art_number`, `read_anc_testing` doesn't read it either.
Also noticed `aggregate_` functions which had a call to `read_csv` which I've removed in favour of the Naomi reading function too.

jeffeaton (approved):

R/input-time-series.R
```diff
@@ -318,9 +318,10 @@ prepare_input_time_series_anc <- function(anc, shape) {
     areas <- shape %>% sf::st_drop_geometry()
   }
 
+  browser()
```
jeffeaton, line 321: Oops, unapproved!

jeffeaton (changes requested):
Ack -- unapprove! There's a browser() call left in there.

R/input-time-series.R
jeffeaton: I've just removed

jeffeaton (approved):

jeffeaton:
I just added a NEWS entry for this version (should have picked that up on review too..oops, sorry).

R/input-time-series.R
r-ash: Oops, good spot - thanks

PR #285, opened 2021-12-10
Update output warnings

rtesra:
This PR adds in a helper function to generate specific warnings for unrealistic outputs simulations. The structure of new warnings can be viewed by running the following test:

https://github.com/mrc-ide/naomi/blob/fa608302cee4ff9caf6578ee15486ab4a8eca657/tests/testthat/test-warning.R#L82

Warning text can be extracted from that test using
```
out$warnings[[1]]$text
out$warnings[[2]]$text
```
Or more generically:
```
w <- capture_condition( naomi_warning("my warning", "model_fit"))
w$text
```


tests/testthat/test-warning.R
```diff
@@ -100,8 +100,8 @@ test_that("warning raised if outputs exceed threshold", {
   expect_length(out$warnings, 2)
   expect_equal(
     out$warnings[[1]]$text,
-    "HIV prevalence is higher than 40% for some age/sex disaggregates")
+    "HIV prevalence is higher than 40% for:  CY2016Q1 MWI_1_1_demo both Y000_004")
```
r-ash, line 103: Do we need to put keys in for this message too? Or will make sense to people without?

R/warnings.R
```diff
@@ -25,3 +25,35 @@ handle_naomi_warnings <- function(expr) {
...
+
+    v <- naomi_output$indicators %>%
+      dplyr::filter(indicator == ind, mean > threshold) %>%
+      dplyr::mutate(disag = paste(calendar_quarter, area_id, sex, age_group, sep = " "))
+
+    if(ind == "prevalence"){t <- t_("WARNING_OUTPUTS_PREV_EXCEEDS_THRESHOLD")}
+    if(ind == "art_coverage"){t <- t_("WARNING_OUTPUTS_ARTCOV_EXCEEDS_THRESHOLD")}
+    strat <- paste0(v$disag, collapse = ", ")
+
+    msg <- paste0(t, " ", strat)
```
r-ash, line 51: Minor comment but traduire can do the second paste0 for you, if your traduire key is e.g.
```
"WARNING_OUTPUTS_PREV_EXCEEDS_THRESHOLD": "HIV prevalence is higher than 40% for: {{rows}}",
```

Can then write the above as
```
if (ind == "prevalence") {
  key <- "WARNING_OUTPUTS_PREV_EXCEEDS_THRESHOLD"
} else if (ind == "art_coverage") {
  key <- "WARNING_OUTPUTS_ARTCOV_EXCEEDS_THRESHOLD"
} else {
  stop("Invalid indicator, can only return warning for prevalence or art_coverage") ## or some generic warning here?
}
msg <- t_(key, list(rows = paste0(v$disag, collapse = ", "))
```

inst/traduire/pt-translation.json
```diff
@@ -309,7 +309,7 @@
     "TYPE_SPECTRUM": "spectrum",
     "TYPE_UNADJUSTED": "Não ajustado",
     "WARNING_OPTIONS_MISSING_ARTATTEND": "Optou por se adaptar ao modelo sem estimar a assistência TARV vizinha. Poderá querer rever a sua selecção para incluir esta opção.",
-    "WARNING_OUTPUTS_PREV_EXCEEDS_THRESHOLD": "A prevalência do VIH é superior a 40% para algumas dissagregações de idade/sexo",
-    "WARNING_OUTPUTS_ARTCOV_EXCEEDS_THRESHOLD": "A cobertura TARV é superior a 100% para algumas dissagregações de idade/sexo",
+    "WARNING_OUTPUTS_PREV_EXCEEDS_THRESHOLD": "A prevalência do VIH é superior a 40% para: ",
```
r-ash, line 312: This is giving a double space `"  "`, one from here and one from the `paste0` `msg <- paste0(t, " ", strat)`

R/warnings.R
```diff
@@ -25,3 +25,35 @@ handle_naomi_warnings <- function(expr) {
...
+
+  if(max(val) > threshold) {
+
+    v <- naomi_output$indicators %>%
+      dplyr::filter(indicator == ind, mean > threshold) %>%
+      dplyr::mutate(disag = paste(calendar_quarter, area_id, sex, age_group, sep = " "))
+
+    if(ind == "prevalence"){t <- t_("WARNING_OUTPUTS_PREV_EXCEEDS_THRESHOLD")}
+    if(ind == "art_coverage"){t <- t_("WARNING_OUTPUTS_ARTCOV_EXCEEDS_THRESHOLD")}
+    strat <- paste0(v$disag, collapse = ", ")
```
r-ash, line 49: How many rows are there potentially? I wonder if we should limit this to say max 20? I think the message will be truncated in the front end anyway but will save sending back loads of data

r-ash (changes requested):
This looks good, the helper function really tidies up some of the logic.

Few comments, some quite minor but would be good to get your responses on at least before merging. I do think we should fix the double space comment though.

tests/testthat/test-warning.R
rtesra: There are so many possible cal quarter/age/sex combinations, I don't think keys will be possible. @jeffeaton do you think read out is clear enough here?

R/warnings.R
rtesra: Thanks - have changed to this (which has dealt with the double space issue as well)

R/warnings.R
rtesra: Potentially this could be big depending on the threshold - but all of the info will be helpful to have locally.

R/warnings.R
```diff
@@ -36,13 +36,24 @@ handle_naomi_warnings <- function(expr) {
...
+    to_upper_first <- function(x) {
+      substr(x, 1, 1) <- toupper(substr(x, 1, 1))
+      x
+    }
+
+
     v <- naomi_output$indicators %>%
       dplyr::filter(indicator == ind, mean > threshold) %>%
-      dplyr::mutate(disag = paste(calendar_quarter, area_id, sex, age_group, sep = " "))
+      dplyr::left_join(naomi_ouput$meta_age_group, by = "age_group") %>%
```
r-ash, line 52: This should be `naomi_output`

r-ash (changes requested):

r-ash (approved):

PR #287, opened 2021-12-13
Make file reading utilities more robust

r-ash:


jeffeaton:
Thanks very much. What will the implication of this be for French formatted files using `,` as decimal mark?

r-ash:
> Thanks very much. What will the implication of this be for French formatted files using , as decimal mark? (And are we supporting that elsewhere in the app? I recall it has come up before, but can't recall the resolution.)

This will fail, but now fail quietly - which isn't ideal. It would have failed before with a parsing error for any numeric column with a `,` in it. This issue came out of Mozambique data which would have been a problem in hintr if the issue was in another column, but it being in `art_new` column hintr happily reads this in as a character and then there is no further validation on its type.

We have no switching of data reading based on the language of file a user has uploaded. We could pick this up from the language they set in the UI. That is a big change in behaviour though and I think all French speaking countries will be using English decimal point . and

r-ash:
Ok - we actually had some code covering this. We look at the delimiter used in the first line, if it is `;` we assume French and use `,` as decimal point. If `,` separated we assume English and use `.` as decimal point.

I've expanded the test - I think this covers all edge cases now

tests/testthat/test-read-data.R
```diff
@@ -59,13 +59,42 @@ test_that("read_*()functions drop NA rows", {
...
 })
+
+test_that("reading utils can handle files with , in numeric columns", {
+  art <- read.csv(a_hintr_data$art_number)
+  art$art_current <- as.character(art$art_current)
+  art$art_current[1] <- "2,031"
+  t <- tempfile()
+  write.csv(art, t)
+  expect_no_error(read_art_number(t))
+})
```
jeffeaton, line 80: Can this test be removed? Looks like the same thing is tested in lines 83-99?

jeffeaton (approved):
Thanks this looks good. I've approved; but with a question -- can the first test be removed as it looks to be tested again in the second test?

tests/testthat/test-read-data.R
r-ash: Oh yes, thanks

PR #293, opened 2022-01-10
increase prevalence threshold

rtesra:
This PR:
* Removes non-outputted indicators from the meta_indicator table in the output package. This addresses issues reported by John Stover that the Spectrum ratio indicators meta data were being written into the output package.
* Adds `anc_tested` and `anc_tested_pos` to anc time series data. This addresses a request from UNAIDs for all data elements informing ANC prevalence calculations to be available in the time series input review plots.
* Increases the threshold to trigger a warning for high HIV prevalence in naomi outputs.

PR #294, opened 2022-01-11
add indicators to anc time series plots

rtesra:
This PR adds the following ANC indicators to the time series plots:
* anc_tested 
* anc_tested_pos 
This addresses a request from UNAIDs for all  data elements informing ANC prevalence calculations to be available in the time series input review plots.


rtesra:
Merged into #293

PR #304, opened 2022-02-18
Update translation of ART in Fr and Pt strings

r-ash:
Ian has asked us to update this to `TARV`


jeffeaton:
I've got a clashing v2.6.14 on PR #303.  Shall we combine these and merge together?

PR #308, opened 2022-03-02
Summary report: update calibration option file type

rtesra:
This PR updates the calibration options read in by the summary report from `calibration_options.csv` -> `calibration_options.yml`

### DIFF
--- a/tests/testthat/test-outputs.R
+++ b/tests/testthat/test-outputs.R
@@ -207,6 +207,14 @@ test_that("can generate summary report from zip file", {
   expect_true(any(grepl("class=\"logo-naomi\"", content)))
 })
 
+test_that("calibration options used in summary report if present", {
+  zip <- hintr_prepare_spectrum_download(a_hintr_output_calibrated)
+  t <- tempfile(fileext = ".html")
+  generate_output_summary_report(t, zip$path, quiet = TRUE)
+  content <- brio::readLines(t)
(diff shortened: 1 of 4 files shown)


r-ash (changes requested):
Can you bump the version number?

tests/testthat/test-outputs.R
```diff
@@ -207,6 +207,14 @@ test_that("can generate summary report from zip file", {
...
 })
 
+test_that("calibration options used in summary report if present", {
+  zip <- hintr_prepare_spectrum_download(a_hintr_output_calibrated)
+  t <- tempfile(fileext = ".html")
+  generate_output_summary_report(t, zip$path, quiet = TRUE)
+  content <- brio::readLines(t)
+  expect_true(any(grepl("Sex and 5-year age group", content)))
+})
+
```
rtesra, line 217: @jeffeaton test here should fail if the subnational age calibration is not printed out in the summary report generated from calibrated results. Did you have something else in mind?

jeffeaton: It did not fail when the `calibration_options` file extension changed—why not? Would like a test that confirms that the printed calibration options come from the `calibration_options.yml` when we expect it to and from `options.yml` when we expect it to. (A test that would have failed and identified the issue that you are fixing in this PR.)

r-ash (approved):

PR #309, opened 2022-03-19
wrap metadata format in quotes to avoid Excel

jeffeaton:
As discussed -- 
* Wraps some of the `format` columns in `metadata.csv` in quotes to avoid Excel mess.
* Removes a single quote when parsed in function `get_metadata()`.
* Adds a regression test for `incidence` indicator format.


tests/testthat/test-metadata.R
```diff
@@ -247,5 +247,10 @@ test_that("metadata format column hasn't been messed by Excel", {
   expect_setequal(meta$format[meta$indicator == "anc_art_coverage"], "0.0%")
   expect_setequal(meta$format[meta$indicator == "recent_infected"], "0.00%")
   expect_setequal(meta$format[meta$indicator == "aware_plhiv_prop"], "0.0%")
+
+  ## Also, 0.0 gets parsed by Excel as 0. To avoid this, 0.0 for the incidence
+  ## indicator is wrapped in quotes. This checks regression against that.
+  expect_setequal(meta$format[meta$indicator == "incidence"], "0.0")
```
r-ash, line 253: You could add a check into this test to make sure that we're including all indicators so that this test will at least fail if we add new indicators in the future
```
expect_seqequal(meta$indicator, c("prevalence", "art_coverage", "incidence", ...))
```

r-ash (changes requested):
This looks good, build has failed because docker image cannot have `/` in its name so let's avoid adding branches with names like this in future. Not a worry for now and I can merge this with build as is (the tests are passing)

r-ash (approved):

PR #311, opened 2022-03-24
Add area hierarchy labels

rtesra:
This PR adds on a area hierarchy column to the data returned by `prepare_input_time_series_art()` and `prepare_input_time_series_anc()`


r-ash:
See #312 which is the cherry picked commits from this so we can separate this from input time series warnings.

PR #296, opened 2022-02-01
add spectrum total warnings

rtesra:

PR #312, opened 2022-03-28
Add area hierarchy labels

r-ash:
Note this is #311 but without the commits of #296. Easier to integrate these independently of each other


r-ash:
We have had countries run at a level below country level before. Do they ever upload data below country level - that would cause issues here as this assumes a single root. We should think about this and add a test if this can happen.

PR #313, opened 2022-04-06
Display counts with a , in large numbers

r-ash:
This will display large numbers instead of e.g. as 12345 as 12,345 as requested by Ian. One thing to note is internationalisation, I think ideally if running app in French or Portuguese we would return `12.345` but elsewhere in the app we are using a `.` for decimal point and that does not get updated when the language is switched. If we need to tidy the language part up we can do that in another ticket.


r-ash:
Tooltip_002
Menu_003
Selection_181

r-ash:
Note this won't work for input time series at the moment. There is a ticket for that and has been added to the backlog.

jeffeaton (approved):
Thanks, this looks good.

PR #318, opened 2022-05-05
Fix anc aggregation

rtesra:
This PR:
* Fixes a bug in the ANC aggregation scripts that was producing duplicate data points in `aggregate_anc()`
* Adds in tests to ensure that distinct strata are maintained in the aggregation and time series scripts


r-ash (approved):
Looks good! Usually good to increase the version number and add news item to a new number but this is fine for now.

PR #327, opened 2022-08-11
Add country specific model & calibration option defaults

r-ash:
This PR will
* Add a function `get_country_option_defaults` to read defaults for a country from yml file. Using yml here because we want to keep the types. It would be nice for editing to have these defaults in a csv but that loses the types of the different options
* Regen docs (fair bit of noise from this)


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: naomi
 Title: Naomi Model for Subnational HIV Estimates
-Version: 2.6.28
+Version: 2.6.29
```
jeffeaton, line 3: Let's bump this to `>2.7.0`

r-ash:
Going to refactor all options stuff into a new package. This no longer needed, see https://github.com/mrc-ide/naomi.options

PR #329, opened 2022-08-11
report mode when simulation error occurs

rtesra:
This PR updates the `extract_indicators()` and `output_package()` function to return mode indicator values when simulating indicators fails. Changes made:
* Added a class to the object returned by `extract_indicators()` to indicate if the mean or mode has been returned
* Added an additional step in `output_package()` that will remove sample if simulation error occurs and generate indicators with only mean values. This will also output a naomi warning that will be displayed to users on the web app to indicate that uncertainty was not generated. 

@jeffeaton At the moment this does not output the original warning from `extract_indicators()` indicating which indicator caused the simulation fail. Options are:
* Output as a warning in naomi scripts - this is useful for debugging purpose
* Output as warning on web app - less useful as it does not give users much of an indication of the source of the problem


jeffeaton:
Code for producing example:

```r
devtools::load_all()

source("~/Downloads/naomi-v2.6.9/tests/testthat/setup-model-frame.R")

a_fit_bad_sample <- a_fit_sample
a_fit_bad_sample$sample$anc_alpha_t2_out[1,] <- NA_real_

out_bad <- output_package(a_fit_bad_sample, a_naomi_data) out_good <-
output_package(a_fit_sample, a_naomi_data)

# For sample with NAs, mode is returned and mean is NA
expect_true("mode" %in% class(out_bad$indicators))
expect_true(all(is.na(out$indicators$mean)))

# For sample without NAs, mode and mode returned expect_true("mean" %in%
class(out_good$indicators))
expect_true(all(!is.na(out_good$indicators$mean)))
```

R/outputs.R
```diff
@@ -183,8 +183,17 @@ extract_indicators <- function(naomi_fit, naomi_mf, na.rm = FALSE) {
...
-  dplyr::select(out, names(naomi_mf$mf_out),
+  out <- dplyr::select(out, names(naomi_mf$mf_out),
                 calendar_quarter, indicator, mean, se, median, mode, lower, upper)
+
+  #  Add a class to indicate if uncertainty has been generated
+  if(!is.null(naomi_fit$sample)) {
+    class(out) <- c("mean", "tbl_df", "tbl","data.frame")
+  } else {
+    class(out) <- c("mode", "tbl_df", "tbl","data.frame")
+  }
```
jeffeaton, line 194: 1. Need more descriptive / specific class names.
2. I would be inclined to put the class name on the output object, rather than the `indicators` table specifically. Is that feasible to do?  Also, let's chat with @r-ash if that is a sensible use of class.

tests/testthat/test-model-fit.R
```diff
@@ -148,6 +148,25 @@ test_that("output_package() works with mode, sample, or both", {
...
 
+test_that("extract_indicators returns mode on error", {
+
+  a_fit_bad_sample <- a_fit_sample
+  a_fit_bad_sample$sample$anc_alpha_t2_out[1,] <- NA_real_
+
+  out_bad <- output_package(a_fit_bad_sample, a_naomi_data)
+  out_good <- output_package(a_fit_sample, a_naomi_data)
+
+  # For sample with NAs, mode is returned and mean is NA
```
jeffeaton, line 159: For this, also test that the warning is generated. (Also will show me how to see the warning :) )

jeffeaton (changes requested):
Thanks for this, this looks generally good to me. 

A comment on the structure: I think that returning `mode` field if no uncertainty generated and `mean` field if is generated will be an undue headache for the front end. In principle, I like keeping this separate, but I suspect it will be far less hassle for @r-ash if we copy the `mode` field into the `mean` field for display purposes.

Let's discuss with @r-ash + also confirm with him that missing values for `lower` / `upper` doesn't create any issues for plotting, etc.

Note to update NEWS and bump version number.

jeffeaton:
@rtesra -- regarding this comment:

> @jeffeaton At the moment this does not output the original warning from extract_indicators() indicating which indicator caused the simulation fail. Options are:
> 
> Output as a warning in naomi scripts - this is useful for debugging purpose
> Output as warning on web app - less useful as it does not give users much of an indication of the source of the problem

I not completely sure what you are referring to with _naomi scripts_? Can you clarify? But I agree it's not so useful to return to the user at the moment.

tests/testthat/test-outputs.R
```diff
@@ -220,8 +220,12 @@ test_that("output_package() catches error if NA in simulated sample.", {
...
   bad_sample <- a_fit_sample
   bad_sample$sample$alpha_t1_out[50] <- NA
-  expect_error(output_package(bad_sample, a_naomi_data),
-               "Error simulating output for indicator: alpha_t1_out. Please contact support for troubleshooting.")
+  out_bad <- output_package(bad_sample, a_naomi_data)
+
+  expect_is(out_bad$indicators, "mode")
+  ## !! TODO: This should still return as a warning, correct?
+  ## expect_warning(<FILL IN HERE>,
+  ##              "Error simulating output for indicator: alpha_t1_out. Please contact support for troubleshooting.")
```
jeffeaton, line 228: @rtesra: This error should still return as a warning, correct? Where does that get attached?

R/outputs.R
rtesra: Have added more specific class names and left the class on the indicators table as discussed

tests/testthat/test-outputs.R
rtesra: Test added in

tests/testthat/test-outputs.R
```diff
@@ -218,14 +218,45 @@ test_that("calibration options used in summary report if present", {
```
rtesra, line 218: @jeffeaton Here is the test and an example of how to extract a naomi warning

jeffeaton: Thanks!

jeffeaton (approved):
Thanks, this looks good to me. Thanks for example on the warnings generation.

I've approved this; I'm holding merging it because I'm not sure if @r-ash is ready for it yet.

rtesra:
hintr build https://github.com/mrc-ide/hintr/pull/329

r-ash:
Run up the app and still some issues in calibrate so this is on hold

PR #332, opened 2022-08-19
Add helper function and test for comparison barchart

r-ash:
This PR will
* Add a function `hintr_comparison_plot` to return data required by hint for plotting comparison barchart
* Adds a test that running this on an old model fit will return a useful error to the user
* Add metadata for comparison barchart
* Regen docs

Note the base branch here, this is merging into return-data-inputs which we can then merge into master


R/outputs.R
```diff
@@ -394,8 +394,10 @@ align_inputs_outputs <- function(naomi_data, indicators, meta_area){
   # Join to get intersecting strata
   join_cols <- c("area_id", "sex", "age_group", "indicator", "calendar_quarter")
 
-  meta_join <- dplyr::semi_join(outputs %>% dplyr::select(join_cols),
-                                inputs_sub %>% dplyr::select(join_cols),
+  meta_join <- dplyr::semi_join(outputs %>%
+                                  dplyr::select(dplyr::all_of(join_cols)),
```
r-ash, line 398: This was to stop dplyr printing some junk when running tests

inst/metadata/colour_scales.csv
```diff
@@ -22,6 +22,8 @@ default,anc_art_new,interpolateOranges,0,350000,FALSE
 default,anc_known_pos,interpolateOranges,0,350000,FALSE
 default,anc_tested_pos,interpolateOranges,0,350000,FALSE
 default,anc_tested_neg,interpolateOranges,0,700000,FALSE
+default,anc_prevalence_age_matched,interpolateOranges,0,0.5,FALSE
```
r-ash, line 25: I think these colour don't actually ever get used as these indicators are only in barchart but we need to add it to satisfy some of the tests

rtesra (approved):

PR #333, opened 2022-08-22
Add default value for births_facility

r-ash:
This is to fix errors seen from the front end when uploading data without a `births_facility` column. If that column is not set we are getting errors from the ANC data here https://github.com/mrc-ide/naomi/pull/330/files#diff-41ca6a0b90b4c5a022737ba75b3a161ccd69eaa2258feb8cb628ed2bd69733bdR283

### DIFF
--- a/R/read-data.R
+++ b/R/read-data.R
@@ -178,8 +178,7 @@ read_anc_testing <- function(file) {
     val[["anc_known_neg"]] <- 0
   }
 
-  if ( !("births_facility" %in% names(val)) ||
-       all(is.na(val[["births_facility"]])) ) {
+  if ( !("births_facility" %in% names(val)) ) {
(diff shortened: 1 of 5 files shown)


R/read-data.R
```diff
@@ -178,6 +178,11 @@ read_anc_testing <- function(file) {
     val[["anc_known_neg"]] <- 0
   }
 
+  if ( !("births_facility" %in% names(val)) ||
+       all(is.na(val[["births_facility"]])) ) {
```
jeffeaton, line 182: This line is redundant since setting default value to `NA` now: `all(is.na(val[["births_facility"]]))`.  

Maybe remove this part of the condition for clarity?

jeffeaton (approved):
This looks good. One optional suggestion for clarity.

PR #337, opened 2022-09-06
Translate metadata in download generation

r-ash:
This PR will
* Refetch the indicator and period metadata when generating output zip so labels are translated

Couple of points here
* We don't currently include sex label in the output table, only the sex values so these are never translated
* We don't currently have any translation for age groups, though the only value which we would need a translation for is "all ages"

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: naomi
 Title: Naomi Model for Subnational HIV Estimates
-Version: 2.7.3
+Version: 2.7.4
 Authors@R: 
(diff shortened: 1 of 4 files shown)


r-ash:
hintr PR https://github.com/mrc-ide/hintr/pull/401

jeffeaton (approved):

jeffeaton:
Discussed 'Couple of points here':

Agreed to create Issues for:
* Include `sex_label` in output table.
* Add translation for `"all ages"` string in `get_age_groups()`

PR #348, opened 2022-10-18
Mrc 3685 additional test

r-ash:
Adding a check that we have metadata for all plot types

There are some plot types which are missing data

We are missing plot metadata for `anc_known_neg` and `births_facility`


inst/traduire/en-translation.json
```diff
@@ -69,6 +71,8 @@
     "INDICATOR_LABEL_ANC_ALREADY_ART": "Number already on ART prior prior to first ANC visit",
     "INDICATOR_LABEL_ANC_ART_NEW": "Number of pregnant women initiated on ART at ANC",
     "INDICATOR_LABEL_ANC_KNOWN_POS": "Number of HIV positive ANC attendees aware of status prior to first ANC visit",
+    "INDICATOR_LABEL_ANC_KNOWN_NEG": "Number of HIV negative ANC attendees recently tested and aware of status prior to first ANC visit",
+    "INDICATOR_LABEL_BIRTHS_FACILITY": "Number of health facility births"
```
r-ash, line 75: Keys have been added "INDICATOR_LABEL_ANC_KNOWN_NEG" and "INDICATOR_LABEL_BIRTHS_FACILITY" but I cannot see where they are used?

rtesra: I though those were used in the front end somewhere (since they exist for other indicators) - are they not required for time series plots?

r-ash: The other `INDICATOR_LABEL_*` strings are used in the `meta_indicator.csv` which I think is what goes into the output zip

rtesra: Ah I see - will remove.

rtesra (approved):

PR #350, opened 2022-10-31
add in home district random effect as default

rtesra:
This PR:
Adds a model option to specify a new structure for ART attending random effects. This change include a random effect for each home district in the ART attending model increasing the "attractiveness" of a home district and preventing too many residents from being reallocated to neighbouring districts. 
* This is now the default model specification with `options$anchor_home_district = TRUE`
* The model can be reverted to the previous specification with `options$anchor_home_district = FALSE`


jeffeaton (changes requested):
The code changes look good to me.

I have edited the NEWS entry to clarify that we have not added any additional random effect parameters to the model, only changed the interpretation of the existing district ART attending random effect parameter. Please review the edited NEWS item and see if it makes sense.

Could we please add some tests for the new specification:
* Confirm that changing the new parameter from TRUE / FALSE as the expected effect on the `tmb_inputs()` result.
* Add an integration test that changing `model_options$anchor_home_district` results in a different output from `hintr_run_model()`.

If it is time sensitive to merge this, I am okay to merge this and then add those tests in a subsequent PR.

jeffeaton:
> The code changes look good to me.
> 
> 
> 
> I have edited the NEWS entry to clarify that we have not added any additional random effect parameters to the model, only changed the interpretation of the existing district ART attending random effect parameter. Please review the edited NEWS item and see if it makes sense.
> 
> 
> 
> Could we please add some tests for the new specification:
> 
> * Confirm that changing the new parameter from TRUE / FALSE as the expected effect on the `tmb_inputs()` result.
> 
> * Add an integration test that changing `model_options$anchor_home_district` results in a different output from `hintr_run_model()`.
> 
> 
> 
> If it is time sensitive to merge this, I am okay to merge this and then add those tests in a subsequent PR.

As discussed on Teams, approving this and we'll start a new PR for tests.

jeffeaton (approved):

PR #351, opened 2022-11-04
mrc-3755: Comparison barchart is showing the wrong indicator for ANC age matched indicators

r-ash:


LekanAnni (changes requested):
Could you add a regression test ?

r-ash:
hintr PR https://github.com/mrc-ide/hintr/pull/418

r-ash:
> Could you add a regression test ?

Added a test which checks that the metadata contains metadata for each indicator only once for each plot type. This would have caught the problem and should catch any similar issues in future.

r-ash:
Screenshot_20221107_114648
This is looking good on dev!

LekanAnni (approved):
LGTM 👍 

Thanks

PR #355, opened 2022-11-14
Add naomi.options to Remotes

paddy7wb:
Add dependency naomi.options to "Imports" in DESCRIPTION, since it is not available on CRAN etc.


paddy7wb:
@r-ash can't seem to add reviewers for this PR so tagging you here.

r-ash (approved):
Looks great, thanks

PR #356, opened 2022-11-17
Vectorise function to get quarter labels

r-ash:
Small change will mean we can speed up some stuff in hintr by using vectorised version!

### DIFF
--- a/R/data-population.R
+++ b/R/data-population.R
@@ -104,15 +104,10 @@ calendar_quarter_to_quarter_id <- function(calendar_quarter) {
 #' @rdname quarter_year_labels
 #' @export
 quarter_id_to_calendar_quarter <- function(quarter_id) {
-  convert <- function(id) {
-    quarter <- id %% 4
-    if (quarter == 0) {
-      quarter <- 4
-    }
-    year <- (id - quarter) / 4 + 1900
-    convert_calendar_quarter(year, quarter)
-  }
-  vapply(quarter_id, convert, character(1))
+  quarter <- quarter_id %% 4
+  quarter[quarter == 0] <- 4
+  year <- (quarter_id - quarter)/4 + 1900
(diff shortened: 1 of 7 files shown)


R/data-population.R
```diff
@@ -104,12 +104,15 @@ calendar_quarter_to_quarter_id <- function(calendar_quarter) {
 #' @rdname quarter_year_labels
 #' @export
 quarter_id_to_calendar_quarter <- function(quarter_id) {
```
jeffeaton, line 106: I think that this will be a much more efficient vectorization than using `vapply()` (and also cleaner to read): 

```
quarter_id_to_calendar_quarter <- function(quarter_id) {

  quarter <- quarter_id %% 4
  quarter[quarter == 0] <- 4
  year <- (quarter_id - quarter)/4 + 1900
  convert_calendar_quarter(year, quarter)
}
```

jeffeaton (changes requested):
Thanks, great. But I think there's a much simpler vectorization of this function. Please see if any reason not to use suggested version.

R/data-population.R
r-ash: That's way better, thanks!

r-ash: Sorry I've been a naughty and shoved a bit of caching code in here to hopefully again make calls from hintr a bit faster.

richfitz (approved):

jeffeaton (commented):
The changes to the date vectorization look good.

Regarding the caching stuff, I haven't reviewed the code at all (thanks Rich). But a question: is there any concern that generic object name `cache` might be prone to inadvertent clashes? Would it be better to give a more specific name, e.g. `naomi_cache`?

Happy if the answer is 'no' and I will approve.

jeffeaton (approved):

PR #357, opened 2022-11-21
mrc-3820: Use qs files to save intermediate hintr output to speed up file read and writes

r-ash:
I'm looking at making some parts of hintr faster. One easy win would be the file format that naomi saves to. At the moment we are saving to rds file. The profile in the screenshot is from hintr reading the calibrate output, and building metadata to return to front end. You can see in profile reading RDS is over half the time ~ 650ms. We could make this faster by using qs files instead https://github.com/traversc/qs

```
> microbenchmark::microbenchmark(readRDS(p), qs::qread(t), times = 10)
Unit: milliseconds
         expr       min        lq      mean    median       uq      max neval
   readRDS(p) 437.01519 450.22436 537.33668 466.87387 656.5683 835.7974    10
qs::qread(t)  80.51279  85.36541  94.03874  87.95275 102.6524 113.6587    10
```

MicrosoftTeams-image (2)

This PR will switch over to save files as qs but keep backwards compatibility so rds output can still be read

### DIFF
--- a/R/inputs-spectrum.R
+++ b/R/inputs-spectrum.R
@@ -435,7 +435,7 @@ read_spectrum_region_code <- function(pjnz) {
 read_spectrum_region_name <- function(pjnz) {
   pjn <- eppasm::read_pjn(pjnz)
   region_name <- pjn[which(pjn[, 1] == "<Projection Parameters - Subnational Region Name2>") + 2, 4]
-  if (isTRUE(region_name == "")) {
+  if (is.null(region_name) || !nzchar(region_name)) {
     region_name <- NA_character_
   }
   region_name


R/outputs.R
```diff
@@ -1276,3 +1276,25 @@ get_period_metadata <- function(calendar_quarters) {
...
+#' This is the data saved from hintr_run_model or hintr_calibrate
+#' before the output zip is generated. This uses the file extension
+#' to identify the reading function to use.
+#'
+#' @param path Path to the file
+#'
+#' @return The read data
+#' @export
+read_hintr_output <- function(path) {
+  is_qs <- grepl("\\.qs$", path, ignore.case = TRUE)
```
jeffeaton, line 1293: there is a builtin function for this: `tools::file_ext()`
```
> tools::file_ext
function (x) 
{
    pos <- regexpr("\\.([[:alnum:]]+)$", x)
    ifelse(pos > -1L, substring(x, pos + 1L), "")
}
```

Not sure there's any benefit to preferring that to `grepl()` though.

inst/report/summary_report.Rmd
```diff
@@ -39,9 +39,9 @@ title: `r title`
 options(scipen=10000000)
 
 # Read files if hintr rds provided
-if(grepl("\\.rds$", params$outputs)) {
+if(grepl("\\.rds|qs$", params$outputs, ignore.case = TRUE)) {
```
jeffeaton, line 42: I don't think this regexp is checking exactly what you want. E.g.
```
> grepl("\\.rds|qs$", "hello.rds_ROB_IS_COOL")
[1] TRUE
```
```
> grepl("\\.rds|qs$", "hello_ROB_IS_COOL_qs")
[1] TRUE
```

I think you want something like:
```
grepl("(\\.rds$)|(\\.qs$)", params$outputs, ignore.case = TRUE)
```

jeffeaton (commented):
Two small comments. One I think should address, but see what you think.

inst/report/summary_report.Rmd
r-ash: Oh thanks, I was trying to be clever here and use grepl instead of tools::file_ext as grepl is faster. But definitely more error prone :wink: and harder to read, and not really *that* much quicker. I've switched these over to use the built in helper.

richfitz: I know rob will change it but you know I love a good regex:
```
grepl("\\.(rds|qs)$", params$outputs, ignore.case = TRUE)
```

jeffeaton (approved):
Thanks! Looks great!

richfitz (approved):
LGTM

R/inputs-spectrum.R
```diff
@@ -435,7 +435,7 @@ read_spectrum_region_code <- function(pjnz) {
 read_spectrum_region_name <- function(pjnz) {
   pjn <- eppasm::read_pjn(pjnz)
   region_name <- pjn[which(pjn[, 1] == "<Projection Parameters - Subnational Region Name2>") + 2, 4]
-  if (region_name == "") {
+  if (isTRUE(region_name == "")) {
```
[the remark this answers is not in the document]
r-ash: I always forget about that one..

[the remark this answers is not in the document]
r-ash: Oh actually - no trying to work around NULL here. Will change the check to be a bit more explicit

PR #360, opened 2022-11-29
population aggregation and pyramid plot initial commit

rtesra:


r-ash:
Not needed, extracting and aggregating this data in hintr and front end

PR #361, opened 2022-12-06
Update model fit to include resource URL for input files

r-ash:
This will allow us to pass in the URL to the resource from the ADR so that we can have a static link to the resource from the ADR saved into the output zip.


jeffeaton (approved):
Looks good to me.

PR #383, opened 2023-02-10
Fix typo

paddy7wb:
`lenth` <- `length` (never used but presume this is a typo? Also, `mean` depends on `prec` and vice versa, so what happens if you don't specify either in this function?)

### DIFF
--- a/R/tmb-model.R
+++ b/R/tmb-model.R
@@ -665,7 +665,11 @@ sample_tmb <- function(fit, nsample = 1000, rng_seed = NULL,
   fit
 }
 
-rmvnorm_sparseprec <- function(n, mean = rep(0, nrow(prec)), prec = diag(lenth(mean))) {
+rmvnorm_sparseprec <- function(
+  n,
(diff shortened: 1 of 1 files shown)


jeffeaton (approved):
Thanks for spotting that!

jeffeaton:
@paddy7wb Thanks for spotting that.

Regarding the function signature, this is lifted from `mvtnorm::rmvnorm()` function:

```
function (n, mean = rep(0, nrow(sigma)), sigma = diag(length(mean)), 
    method = c("eigen", "svd", "chol"), pre0.9_9994 = FALSE, 
    checkSymmetry = TRUE) 
```

If you don't specify either, bad things happen:
```
> mvtnorm::rmvnorm(3)
Error in nrow(sigma) : 
  promise already under evaluation: recursive default argument reference or earlier problems?
```

But specifying the defaults that way allows you to specify only one or the other, which is a convenient usage to be able to give it a mean vector and standard IID precision, or give it a precision matrix and default to a mean 0 process.

PR #387, opened 2023-02-27
Render comparison plot when ANC T1 != model T1

rtesra:
Quick fix to prevent ANC plots from rendering if ANC T1 is not in the same year as model T1
TO DO: Refactor align_inputs_outputs script to align based on data inclusion vs. matching on years


jeffeaton (approved):
Thanks — looks good to me.

r-ash (approved):

PR #391, opened 2023-03-17
Explictly state spdep::nb2listw style argument, to avoid warning

potoole7:
Actually using `spdep::mat2listw`, not `spdep::nb2listw`!


r-ash:
Addressed in https://github.com/mrc-ide/naomi/pull/418

PR #402, opened 2023-03-28
False convergence warning to error

rtesra:


r-ash (approved):
This looks good to me, can you bump the version number?

PR #404, opened 2023-05-16
Reduce number of indicators returned in plotting data

r-ash:
We've seen the plotting data returned to the front end for DRC growing too big and it causing issues with transfer and serializing times. Serializing to json taking about 35s and data transfer can take several minutes which is causing problems for people on slower connections. The serialized json is going to 600MB which is too big for a single response. This PR puts in a temporary fix to cut down the number of indicators returned, eventually we will split this data transfer up instead to send a request to get plotting data for each indicator/time separately. But this is a quick fix for now.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: naomi
 Title: Naomi Model for Subnational HIV Estimates
-Version: 2.9.10
+Version: 2.9.11
 Authors@R:
     person(given = "Jeff",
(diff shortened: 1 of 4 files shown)


r-ash:
hintr PR https://github.com/mrc-ide/hintr/pull/465

rtesra (approved):

r-ash:
DRC are unblocked by using firefox so we don't need this temporary hack anymore!

PR #405, opened 2023-08-21
mrc-4555: Add ability to save hintr output as duckdb database

r-ash:
This PR will
* Add the ability to save hintr output as a duckdb database, based upon the file extension.

This will be used from hint to slice and read smaller portions of the data so we can deal better with large datasets (like for DRC)

### DIFF
--- a/R/Lproj.R
+++ b/R/Lproj.R
@@ -254,7 +254,8 @@ create_Lproj <- function(spec, mf_model,
     dplyr::inner_join(
       dplyr::select(mf_model, spectrum_region_code, sex, age_group_infection = age_group, area_id, idx1 = idx),
       by = c("spectrum_region_code", "sex", "age_group_infection"),
-      multiple = "all"
+      multiple = "all",
+      relationship = "many-to-many"
     ) %>%
     dplyr::inner_join(
       dplyr::select(mf_model, spectrum_region_code, sex, age_group2 = age_group, area_id, idx2 = idx),
--- a/R/Lproj.R
+++ b/R/Lproj.R
@@ -197,7 +197,8 @@ create_Lproj <- function(spec, mf_model,
     dplyr::inner_join(
       dplyr::select(mf_model, spectrum_region_code, sex, age_group1 = age_group, area_id, idx1 = idx),
       by = c("spectrum_region_code", "sex", "age_group1"),
-      multiple = "all"
+      multiple = "all",
+      relationship = "many-to-many"
     ) %>%
     dplyr::inner_join(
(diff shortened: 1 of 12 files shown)


R/Lproj.R
```diff
@@ -197,7 +197,8 @@ create_Lproj <- function(spec, mf_model,
     dplyr::inner_join(
       dplyr::select(mf_model, spectrum_region_code, sex, age_group1 = age_group, area_id, idx1 = idx),
       by = c("spectrum_region_code", "sex", "age_group1"),
-      multiple = "all"
+      multiple = "all",
+      relationship = "many-to-many"
```
r-ash, line 201: New warnings coming from dplyr related to many-to-many relationships. See https://www.tidyverse.org/blog/2023/03/dplyr-1-1-1/#revisiting-multiple-matches . I've added these where CI was throwing a warning. @rtesra can you check that these are where you expect them to be?

R/outputs.R
```diff
@@ -1322,10 +1326,21 @@ get_period_metadata <- function(calendar_quarters) {
 #' @return The read data
 #' @export
 read_hintr_output <- function(path) {
-  if (tolower(tools::file_ext(path)) == "qs") {
+  type <- tolower(tools::file_ext(path))
+  if (type == "qs") {
     qs::qread(path)
+  } else if (type == "duckdb") {
+    read_duckdb(path)
```
richfitz, line 1333: with three options here now, consider explicitly adding `type == "rds"` and then a final `else stop(...)` branch

tests/testthat/test-hintr-plot-data.R
```diff
@@ -52,3 +52,22 @@ test_that("there is metadata for every indicator in comparison data", {
...
 })
+
+test_that("hintr data can be saved and read as qs or duckdb type", {
+  t_qs <- tempfile(fileext = ".qs")
+  t_db <- tempfile(fileext = ".duckdb")
+  d <- data.frame(x = c(1, 2, 3), y = c(4, 5, 6))
+
+  hintr_save(d, t_qs)
+  hintr_save(d, t_db)
+  expect_equal(read_hintr_output(t_qs), read_hintr_output(t_db))
```
M-Kusumgar, line 63: should we not test each case against the concrete dataframe because if there is a similar bug in both in the future then this test could still pass

richfitz (approved):

M-Kusumgar (approved):

R/Lproj.R
jeffeaton: Thanks — I've just reviewed these. All look good and as expected.

jeffeaton (approved):

rtesra (approved):

PR #406, opened 2023-09-06
Set T4 and T5 projection quarters when missing from model options

rtesra:
This PR fixes a bug in `naomi_prepare_data()` when reading in regressed model options where T4 and T5 were not specified. 

New behaviour to automatically set T4 and T5 to 24- and 36-months ahead of T3 if they are missing from model options.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# naomi 2.9.11
+
+* Automatically set T4 and T5 to 24- and 36-months ahead of T3 if not specified in model options
+
 # naomi 2.9.10
 
 * Update PJNZ extraction for adult ART need Dec 31 for 2023 PJNZ files. Previously child ART was
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # naomi 2.9.11
 
-* Automatically set T4 and T5 to 24- and 36-months ahead of T3 if not specified in model options
+* Automatically set T4 and T5 to 24- and 36-months ahead of T3 if not specified in model options.
 
 # naomi 2.9.10
 
--- a/R/run-model.R
+++ b/R/run-model.R
@@ -328,7 +328,7 @@ naomi_prepare_data <- function(data, options) {
 
   if(is.null(options$calendar_quarter_t5)) {
     # If T5 is not specified, set T4 to 36-months after T3
-    t5 <- calendar_quarter_to_quarter_id(calendar_quarter_t3) + 9
+    t4 <- calendar_quarter_to_quarter_id(calendar_quarter_t4) + 3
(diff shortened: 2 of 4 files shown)


R/run-model.R
```diff
@@ -317,8 +317,23 @@ naomi_prepare_data <- function(data, options) {
...
+    # If T4 is not specified, set T4 to 24-months after T3
+    t4 <- calendar_quarter_to_quarter_id(calendar_quarter_t3) + 6
+    calendar_quarter_t4 <- quarter_id_to_calendar_quarter(t4)
+  } else{
+    calendar_quarter_t4 <- options$calendar_quarter_t4
+  }
+
+  if(is.null(options$calendar_quarter_t5)) {
+    # If T5 is not specified, set T4 to 36-months after T3
+    t5 <- calendar_quarter_to_quarter_id(calendar_quarter_t3) + 9
```
jeffeaton, line 331: Thanks—in a pathological situation, this could create an issue if T4 was set but T5 was not and T4 was more than 9 months after T3.

That _shouldn't_ ever happen.  But maybe just for completeness set this 3 months after T4?

(This would still leave open a potential issue in a situation where T5 was set to a time less than 6 months after T3, but T4 was not set. But maybe just leave that as an error.)

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# naomi 2.9.11
+
+* Automatically set T4 and T5 to 24- and 36-months ahead of T3 if not specified in model options
```
jeffeaton, line 3: Add period.

jeffeaton (commented):
Thanks — looks good. I have made one suggestion; see if it makes sense.

Thanks,
Jeff

R/run-model.R
rtesra: @jeffeaton That makes sense - have set T5 relative to T4.

R/run-model.R
```diff
@@ -328,7 +328,7 @@ naomi_prepare_data <- function(data, options) {
 
   if(is.null(options$calendar_quarter_t5)) {
     # If T5 is not specified, set T4 to 36-months after T3
-    t5 <- calendar_quarter_to_quarter_id(calendar_quarter_t3) + 9
+    t4 <- calendar_quarter_to_quarter_id(calendar_quarter_t4) + 3
```
jeffeaton, line 331: Thanks — I think you have an error here though; should `t4 <- ...` be `t5 <- ...`?

Also, does the test catch this?

jeffeaton (changes requested):
Thanks for updating. I think there might be a typo though — see comment.

r-ash (approved):

jeffeaton (approved):

PR #411, opened 2023-10-18
Add dummy AGYW download

r-ash:
This PR will add a dummy xlsx download for AGYW. Adding this so we can build the hintr and web app parts out


R/car.R
```diff
@@ -74,7 +74,7 @@ scale_gmrf_precision <- function(Q,
                                  A = matrix(1, ncol = ncol(Q)),
                                  eps = sqrt(.Machine$double.eps)) {
 
-  nb <- spdep::mat2listw(abs(Q))$neighbours
+  nb <- spdep::mat2listw(abs(Q), style = "M")$neighbours
```
r-ash, line 77: This was raising a new warning in hintr tests

```
Warning (test-03-run-model.R:185:5): real model can be run
--
  | style is M (missing); style should be set to a valid value
  | Backtrace:
  | ▆
  | 1. ├─withr::with_envvar(...) at test-03-run-model.R:184:2
  | 2. │ └─base::force(code)
  | 3. └─hintr:::run_model(data, options, tempdir()) at test-03-run-model.R:185:4
  | 4.   └─naomi::hintr_run_model(data, options, model_output_path, validate = FALSE) at src/R/run_model.R:75:2
  | 5.     ├─naomi:::handle_naomi_warnings(run_model(data, options, validate))
  | 6.     │ ├─base::withCallingHandlers(...)
  | 7.     │ └─base::force(expr)
  | 8.     └─naomi:::run_model(data, options, validate)
  | 9.       └─naomi:::naomi_prepare_data(data, options)
  | 10.         └─naomi::naomi_model_frame(...)
  | 11.           └─naomi::scale_gmrf_precision(...)
  | 12.             └─spdep::mat2listw(abs(Q))

```

I think this is right to set style to `M`? From docs it says for `style` arg "default NULL, missing, set to "M" and warning given; if not "M", passed to [nb2listw](http://127.0.0.1:40083/help/library/spdep/help/nb2listw) to re-build the object" so I think if it is missing then will default to "M" but let me know if this is incorrect

rtesra (approved):

PR #413, opened 2023-11-01
SHIPP tool integration

rtesra:
To do:

- [x] Add tests
- [x] Add function to update excel workbook with outputs from `agyw_format_naomi()`
- [x] Add scale to KP consensus estimates from KP workbook
- [x] Add scale to new infections estimates from KP workbook or Goals ASM database
- [x] Test workbook on multiple countries


R/agyw-integration.R
```diff
@@ -0,0 +1,1475 @@
...
+#' @param male_logit_prevalence Risk adjusted estimates of male prevalence in sexual risk groups generated by `agyw_calculate_prevalence_male()`
+#' @param survey_year NOTES:: add in description + when this should be adjusted. Hardcoded to 2018.
+#'
+#' @return Wide format output required for the AGYW workbook
+#'
+#' @export
+
+
+
+agyw_generate_risk_populations <- function(naomi_output) {
```
rtesra, line 1406: @r-ash this functions takes in a naomi output zip or hintr object and retursn the three data.frames needed to update the agyw tool

R/car.R
```diff
@@ -74,7 +74,7 @@ scale_gmrf_precision <- function(Q,
                                  A = matrix(1, ncol = ncol(Q)),
                                  eps = sqrt(.Machine$double.eps)) {
 
-  nb <- spdep::mat2listw(abs(Q), style = "M")$neighbours
+  nb <- spdep::mat2listw(abs(Q), style = "B")$neighbours
```
r-ash, line 77: Note I've updated this using Paddy's fix from https://github.com/mrc-ide/naomi/pull/391 which should stop the huge amount of warnings. We can close that PR after this has been merged

r-ash: Actually tried that and it broke everything, just suppressing this warning on "M" for now as I can't find an equivalent to this which won't trigger a warning

R/agyw-integration.R
```diff
@@ -0,0 +1,1513 @@
...
+    dplyr::select(iso3, area_id, total_fsw, age_group, area_level, spectrum_region_code)
+
+  #' Check for consensus estimate of FSW
+  kp_consensus <- extract_kp_workbook(pjnz)
+  fsw_consensus <- kp_consensus[kp_consensus$key_population == "FSW", ]$population_size
+
+  if(!is.na(fsw_consensus)){
+
+    # Check if consensus estimate is larger than age matched population denominator
+    pop <- naomi_pop[naomi_pop$area_level == 0 & naomi_pop$age_group == "Y015_049" & naomi_pop$sex == "female",]$population
```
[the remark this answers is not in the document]
osymandius: Would be happy to throw an error (or use the default admin1 PSE% from our analysis) in cases where consensus estimates are >10% of the sex/age matched denominator.

R/agyw-integration.R
```diff
@@ -0,0 +1,1517 @@
...
+#' @param male_logit_prevalence Risk adjusted estimates of male prevalence in sexual risk groups generated by `agyw_calculate_prevalence_male()`
+#' @param survey_year NOTES:: add in description + when this should be adjusted. Hardcoded to 2018.
+#'
+#' @return Wide format output required for the AGYW workbook
+#'
+#' @export
+#'
+#' Survey year should be updated to most current household survey in the country -
+#' for countries without recent household surveys, leave at 2018 - the spatiotemporal
+#' model of sexual behaviour fitted to all countries has the most data for in roughly 2018
```
r-ash, line 1434: Can you fix the docs here, couple of things
* Update the parameters
* The details about "survey year" should either be under the parameter, after the title or under a `@details` section. This won't go into the generated docs where it is atm
* Does this need to be exported? I think it is only used internally in which case replace the `export` with `@keywords internal`

tests/testthat/helper-testing.R
```diff
@@ -3,6 +3,8 @@ expect_no_error <- function(object) {
 }
 
 with_mock <- function(..., .parent = parent.frame()) {
+  ## Don't use this, this should be removed, now testthat have
+  ## added with_mocked_bindings that should be preferred
```
r-ash, line 7: Thanks for adding this!

R/agyw-integration.R
```diff
@@ -0,0 +1,1517 @@
+#' Format naomi outputs for PSE tool
+#'
+#' @param outputs Naomi output
+#' @param options Naomi model options.
+#'
+#'
+#' @return Naomi indicators formatted for the AGYW workbook.
+#'
+#' @export
```
r-ash, line 9: Does this actually need to be exported? Will you use it from anywhere other than Naomi?

R/agyw-integration.R
```diff
@@ -0,0 +1,1517 @@
...
+#'
+#' @param outputs Naomi output.
+#' @param options Naomi model options.
+#' @param naomi_population Naomi population estimates for T2.
+#' @param kp_consensus Key pop consensus estimates.
+#'
+#'
+#' @return District level FSW estimates by 5-year age bands for ages 15-49.
+#'
+#' @export
```
r-ash, line 163: Same with this export?

R/agyw-integration.R
```diff
@@ -0,0 +1,1517 @@
...
+
+
+agyw_calculate_prevalence_male <- function(naomi_output,
+                                           areas,
+                                           options,
+                                           msm_est,
+                                           male_srb,
+                                           survey_year_sample = 2018) {
+
+  #' Naomi estimates of PLHIV and population by district and age band
```
r-ash, line 851:
```suggestion
  # Naomi estimates of PLHIV and population by district and age band
```

r-ash: And elsewhere in this file

R/agyw-integration.R
```diff
@@ -0,0 +1,1517 @@
...
+                                      c(0, 0.3, 1, 3, 10^6),
+                                      labels = c("Low", "Moderate", "High", "Very High"),
+                                      include.lowest = TRUE, right = TRUE))
+
+  #' Check that sum of disaggregated infections is the same as total infections
+  sum_infections <- df3$infections_nosex12m + df3$infections_sexcohab + df3$infections_sexnonreg + df3$infections_sexpaid12m
+
+  # TO DO: Flag this to add in warning - stop please contact support (usually an issue with mapping boundaries)
+  # ADD IN WARNIING HERE
+  stopifnot(max(df3$infections - sum_infections) < 10^{-9})
```
r-ash, line 1186: Do you want to do this as part of this PR? stopifnot's generally good to avoid as the error message they produce is not easy to understand by users

r-ash (changes requested):
This looks good, couple of notes about `#'` being used outside of function docs and use of `export`. I've not looked very deeply at the data manipulation stuff, agree that we should probably run this for a few countries and see if it highlights any edge cases we've not thought about.

R/car.R
```diff
@@ -74,7 +74,17 @@ scale_gmrf_precision <- function(Q,
                                  A = matrix(1, ncol = ncol(Q)),
                                  eps = sqrt(.Machine$double.eps)) {
 
-  nb <- spdep::mat2listw(abs(Q), style = "M")$neighbours
+  # spdep is annoying warning with style = "M" but not clear from docs
+  # if this means style missing or style Matrix. And none of the other
+  # options for style seemingly do the same thing as style = "M"
+  # so just ignoring this warning for now
+  withCallingHandlers({
+    nb <- spdep::mat2listw(abs(Q), style = "M")$neighbours
```
r-ash, line 82: There are new warnings coming from this.

In the current build

```
Warning (test-car.R:12:3): scale_gmrf_precision() returns expected values
6 is not an origin
Backtrace:
    ▆
 1. └─naomi::scale_gmrf_precision(Q) at test-car.R:12:3
 2.   ├─base::withCallingHandlers(...) at src/R/car.R:81:3
 3.   └─spdep::mat2listw(abs(Q), style = "M") at src/R/car.R:82:5
 4.     └─spdep::sn2listw(...)
Warning (test-car.R:12:3): scale_gmrf_precision() returns expected values
no-neighbour observations found, set zero.policy to TRUE;
this warning will soon become an error
Backtrace:
    ▆
 1. └─naomi::scale_gmrf_precision(Q) at test-car.R:12:3
 2.   ├─base::withCallingHandlers(...) at src/R/car.R:81:3
 3.   └─spdep::mat2listw(abs(Q), style = "M") at src/R/car.R:82:5
 4.     └─spdep::sn2listw(...)
```

Is there anything we can do about this? Can we replace this function entirely? It feels like this package is always chucking out warnings and there is a new one every few weeks.

R/agyw-integration.R
rtesra: Thanks - have removed.

R/agyw-integration.R
rtesra: Thanks - have removed.

R/agyw-integration.R
rtesra: Thanks - have updated.

R/agyw-integration.R
rtesra: Have converted into more informative warning.

R/car.R
rtesra: @jeffeaton I thank we have discussed this in the past - any thoughts?

jeffeaton: just confirming this is resolved now in #418.

R/agyw-integration.R
```diff
@@ -88,65 +88,64 @@ agyw_format_naomi <- function(outputs, options){
...
-                                             "The Gambia" = "GMB",
-                                             "Guinea bissau" = "GNB",
-                                             "Equatorial Guinea" = "GNQ",
-                                             "Niger" = "NER",
-                                             "Nigeria" = "NGA",
-                                             "Senegal" = "SEN"))
-
-
 
+  country_name_db <- tibble::tribble(~country, ~iso3,
```
rtesra, line 98: @krisher1 I have updated the method for hardcoding the country names to pulling straight from a dataframe.

R/agyw-integration.R
```diff
@@ -192,9 +191,7 @@ agyw_format_naomi <- function(outputs, options){
   if(options$area_scope=="AGO") {
     naomi_wide$area_name <- stringr::str_to_title(naomi_wide$area_name)
   }
-  if(options$area_scope=="COD") {
-    naomi_wide$Country <- "Democratic Republic of the Congo"
-  }
+
```
rtesra, line 194: @krisher1 the change in code above should maintain the correct name for DRC.

R/agyw-integration.R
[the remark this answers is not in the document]
jeffeaton: Will defer. 

What will happen if this fails though? I see in next line that it has a `stopifnot()`.  Should we manage the fail more gently?

R/agyw-integration.R
```diff
@@ -0,0 +1,1638 @@
...
+        relative_prop = total_pwid/sum(total_pwid),
+        consensus_pse = pwid_consensus,
+        total_pwid = consensus_pse * relative_prop)
+
+    pwid <- pwid_scaled %>% dplyr::select(-consensus_pse, relative_prop)
+  }
+
+
+  # Assumption from literature that 9% of PWID are female so remove them from
+  # the male denominator
```
jeffeaton, line 416: @osymandius: can we add any reference in the comment here to the 'literature' this is from?

R/agyw-integration.R
```diff
@@ -0,0 +1,1638 @@
...
+
+  pwid$total_pwid <- pwid$total_pwid * 0.91
+
+  # PWID age distribution
+  # Review of literature - Hines et al Lancet Global Health 2020
```
jeffeaton, line 421: Thanks — add DOI or link if possible

R/agyw-integration.R
```diff
@@ -0,0 +1,1638 @@
...
+  # Calculate risk group incidence for aggregate age groups
+
+  summarise_age_cat_female <- function(dat, age_cat) {
+
+    if (age_cat == "Y015_024") {age_groups <- c("Y015_019", "Y020_024")}
+    if (age_cat == "Y025_049") {age_groups <- c("Y025_029","Y030_034","Y035_039",
+                                                "Y040_044", "Y045_049")}
+    if (age_cat == "Y015_049") {age_groups <- c("Y015_019", "Y020_024","Y025_029",
+                                                "Y030_034","Y035_039","Y040_044",
+                                                "Y045_049")}
```
jeffeaton, line 1186: Similar comments to above; would prefer to avoid hardcoding these; fine to leave for now.

R/downloads.R
```diff
@@ -115,8 +115,18 @@ hintr_prepare_agyw_download <- function(output, pjnz,
...
+
+  risk_populations <- agyw_generate_risk_populations(output$model_output_path,
+                                                     pjnz)
+
+  sheets <- list(
+    "All outputs - F" = risk_populations$female_incidence,
+    "All outputs - M" = risk_populations$male_incidence,
+    "NAOMI outputs" = risk_populations$naomi_output
+  )
+  write_xlsx_sheets(template_path, sheets, path = path)
```
jeffeaton, line 129: What happens if something fails / throws an error in here?

man/agyw_adjust_sexbehav_fsw.Rd
jeffeaton, line 1: Do all of these functions need to be exported / have `man` files generated?  I see functions labelled as `@internal`.  I think if you add something like `@noRd` it will also suppress generating the `man` files (@r-ash can confirm).  Or maybe it's an intentional choice to generate them.

jeffeaton (commented):
Thanks @rtesra and all. This is heroic.  Also clear and well commented, thanks.

I haven't checked details (obviously), but have reviewed enough that not concerned this will interact poorly with anything else.

I have left some styling comments in here. You can basically take / add to an issue for later / leave forever those at your discretion.

Two of the comments that I would like check: 

1. What will happen if `hintr_prepare_agyw_download()` fails? [What will user see?]
2. Are we intentionally generating `man` files for all the functions here?  I think can add `@noRd` to the Roxygen to suppress.

R/agyw-integration.R
```diff
@@ -0,0 +1,1638 @@
...
+    dat %>%
+      dplyr::select(area_id, area_name, area_level, spectrum_region_code, calendar_quarter,
+                    age_group, sex, indicator, mean) %>%
+      tidyr::pivot_wider(names_from = indicator, values_from = mean) %>%
+      dplyr::group_by(area_id, area_name, area_level,spectrum_region_code,
+                      calendar_quarter, sex) %>%
+      dplyr::summarise(
+        "population" = sum(population * as.integer(age_group %in% age_groups)),
+        "plhiv" = sum(plhiv * as.integer(age_group %in% age_groups)),
+        "infections" = sum(infections * as.integer(age_group %in% age_groups)),
```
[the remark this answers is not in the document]
rtesra: These aggregation do not exist in the indicators currently (15-29 and 25-49). The purpose of this function is formate and aggregate Naomi inputs for the rest of of the workflow - will add some additional context in in the functions description.

R/agyw-integration.R
```diff
@@ -0,0 +1,1638 @@
...
+
+  # Naomi indicators for aggregate age groups
+  df1 <- dplyr::bind_rows(summarise_naomi_ind(naomi_ind_labelled, "Y015_024"),
+                          summarise_naomi_ind(naomi_ind_labelled, "Y025_049"))
+
+  # Naomi indicators for 5-year age groups + 15-49
+  df2 <- naomi_ind_labelled %>%
+    dplyr::filter(age_group %in% c("Y015_019", "Y020_024", "Y025_029", "Y030_034",
+                                   "Y035_039", "Y040_044", "Y045_049", "Y015_049")) %>%
+    dplyr::select(names(df1)) %>%
```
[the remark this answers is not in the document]
rtesra: Not following on how we would select these age groups using a join?

R/agyw-integration.R
```diff
@@ -0,0 +1,1638 @@
...
+                                     "Zimbabwe", "ZWE",
+                                     "Angola", "AGO",
+                                     "Burundi", "BDI",
+                                     "Democratic Republic of the Congo", "COD",
+                                     "Gabon", "GAB",
+                                     "Rwanda", "RWA",
+                                     "Ethiopia", "ETH",
+                                     "Haiti", "HTI",
+                                     "Chad", "TCD",
+                                     "Cote D'Ivoire", "CIV",
```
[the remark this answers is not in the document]
rtesra: @krisher1 I think that is for you - am I correct in thinking this is hardcoded to match the workbooks?

R/agyw-integration.R
rtesra: Added here: https://github.com/mrc-ide/naomi/pull/413/commits/e923fceed2f0c943e3a3a6423fb6e9407bdd7d73

man/agyw_adjust_sexbehav_fsw.Rd
r-ash: `@keywords internal` removes the function from the documentation index, `@noRd` means the roxygen docs aren't generated at all.

Looks like people to tend to use `@noRd` for internal functions, and `@keywords internal` for functions aimed at other developers, not typical users of the package.

So yep I agree, `@noRd` is better here.

R/agyw-integration.R
```diff
@@ -0,0 +1,1659 @@
...
+  fsw_consensus <- kp_consensus[kp_consensus$key_population == "FSW", ]$population_size
+
+  if(!is.na(fsw_consensus)){
+
+    # Check if consensus estimate is larger than age matched population denominator
+    pop <- naomi_pop[naomi_pop$area_level == 0 & naomi_pop$age_group == "Y015_049" & naomi_pop$sex == "female",]$population
+    prop_pop <- fsw_consensus / pop
+
+    if(prop_pop >= 0.05) {
+      t_("FSW_CONSENSUS_WARNING")
```
r-ash, line 254: This call doesn't go anywhere? Did you mean it as a waning

R/agyw-integration.R
```diff
@@ -0,0 +1,1659 @@
...
+  # Check for consensus estimate of MSM
+  pwid_consensus <- kp_consensus[kp_consensus$key_population == "PWID", ]$population_size
+
+  if(!is.na(pwid_consensus)){
+    # Check if consensus estimate is larger than age matched population denominator
+    pop <- naomi_pop[naomi_pop$area_level == 0 & naomi_pop$age_group == "Y015_049" & naomi_pop$sex == "male",]$population
+    prop_pop <- pwid_consensus / pop
+
+    if(prop_pop >= 0.05) {
+      t_("PWID_CONSENSUS_WARNING")
```
r-ash, line 409: This also doesn't go anywhere

r-ash:
> On warnings in general, Ashton, Robert we should add a clearer error for if the workbook fails to generate. So far we have the following warnings/errors in place:
Error: If area hierarchy if Naomi inputs does not match any of external data inputs. Flags which area_ids are inconsistent and tells user to contact support.
Warning: (I'll add this in now) If KP consensus estimate is larger than 5% of matched population denominator. Tells user that default proportions will be used instead. 
Error: If sum of disaggregated infections does not equal total infections - tells user to contact support. This shouldn't happen now that we flag the mismatched area_ids at the start but left it for edge cases (who knows).

Errors sounds good, I will add some specific error handling into the excel file generation to return a nicer message should this fail.

Note about the warnings, these are not surfaced at the moment for downloads. So fine to add these here if you want but users won't see them until we add some specific handling for this in the front end. It will require a new endpoint in hintr to get these surfaced.

In general also after seeing the other warnings I am reluctant to use them. I think people have been trained to ignore the warnings.

That said, see the note above about a few floating calls to `t_` which I think you want to wrap in a warning?

R/downloads.R
r-ash: Thanks, I've written a wrapper which will add some more context should anything in excel writing error. Should return some more info to users https://github.com/mrc-ide/naomi/pull/413/commits/a10bfcd0a8f60407c69acf400e402f42736afb25#diff-8e11e9d56ce8b4fdd117a811bd64a4b2ea57306740475e9703d27faa1131cac3R1665

R/agyw-integration.R
rtesra: Have removed and fixed up test to check that this working.

R/agyw-integration.R
rtesra: Have removed and fixed up test to check that this working.

R/agyw-integration.R
rtesra: @osymandius and @krisher1 following up on this?

PR #418, opened 2023-12-11
Add VMMC to spectrum download

r-ash:
Just a stub atm, impl TODO

This PR will
* Add vmmc_path as an argument to spectrum download file, default to null

@jeffeaton will do the work to read this file and add it to the relevant file in the output zip


R/agyw-integration.R
jeffeaton, line 1: Is this file intended to be in this PR? Should it rather be in #413?

R/outputs.R
```diff
@@ -977,8 +980,17 @@ save_output <- function(filename, dir,
...
 
   if (export_datapack) {
+
+    if (!is.null(vmmc_path)) {
+      assert_scalar_character(vmmc_path)
+      ## Skip the first row, the file has two rows of headers
+      vmmc_datapack <- openxlsx::read.xlsx(vmmc_path, sheet = "Datapack inputs",
+                                           startRow = 2)
+      ## TODO: Add it to relevant place in download
+    }
```
jeffeaton, line 990: @r-ash -- note that I added a commit to move this read file inside the `if (export_datapack)`.

Let me know if any issue with that.

R/agyw-integration.R
r-ash: Not intended to be here, thanks for spotting that.

R/pepfar-datapack.R
```diff
@@ -1,3 +1,5 @@
+PEPFAR_DATAPACK_PATH <- "pepfar_datapack_indicators_2024.csv"
```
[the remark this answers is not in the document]
r-ash: Just because I wanted to reference the same path from a test, so now if we change the path for any reason we'll only have to do that in one place. Instead of in the code and in the test.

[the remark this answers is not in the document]
jeffeaton: Thanks, sounds good. I added a commit a7a51252: 
* change `PEPFAR_DATAPACK_PATH` -> `PEPFAR_DATAPACK_FILENAME`
* add a comment in `outputs.R` to note it is defined in `pepfar-datapack.R`

Let me know if that all okay.

jeffeaton (approved):
Thanks, this looks good. I have additionally: 

* Added a commit changing `PEPFAR_DATAPACK_PATH` -> `PEPFAR_DATAPACK_FILENAME`; let me know if okay
* Merged #423 addressing issue #421 into here and closed #423

R/outputs.R
```diff
@@ -977,8 +980,17 @@ save_output <- function(filename, dir,
   }
 
   if (export_datapack) {
+
+    if (!is.null(vmmc_path)) {
+      assert_scalar_character(vmmc_path)
+      ## Skip the first row, the file has two rows of headers
+      vmmc_datapack <- openxlsx::read.xlsx(vmmc_path, sheet = "Datapack inputs",
+                                           startRow = 2)
```
jeffeaton, line 988: How do we want this to fail if sheet is not found?

Currently it will return an error from `openxlsx`: `Cannot find sheet named "Datapack inputs"`

r-ash: Good spot, just added a better check to hintr for this case. So users will see an error on file upload. See https://github.com/mrc-ide/hintr/pull/485/commits/5a56a2d31b10dbcc0624e4f3bc0d5544d077cf6b

So with that, we could make this check better in the same vein as the hintr commit. So this message will be unlikely to show for users of the web app now, but will be shown if running it via R.

r-ash: Will leave this for addressing in #424

PR #427, opened 2024-01-24
Speed up ART time series generation

r-ash:
This PR will speed up generation of ART time series.

We've seen with current Malawi data that this is taking up to 25s to return the data 

image

Running a profile we can see that a lot of this is coming from the data prep, with the json serialization not being insignificant

image

This PR does a couple of things
* Use new `|>` pipe instead of `%>%`, we should in general switch over to this as `%>%` does incur a small overhead whereas `|>` does not
* Don't split the data and loop over it in aggregation script, instead rely on the group by to manage this splitting for us.

I've checked this with the current Malawi data and it produces the same result, in a slightly different order. But we should check this carefully and make sure it will produce the same output for other countries too. Hopefully the tests will check the edge cases for us here.

This speeds up `aggregate_art` quite a lot, before this was taking ~12s on my machine, after this change taking ~850ms so a decent speedup! 

I think worth having a closer look at this at some point but hopefully this should help for now.

If you want to run a profile yourself `profvis` is a good way to go

```
profvis::profvis(aggregate_art(art, shape), simplify = FALSE)
```


r-ash:
Great tests here! 

So my changes collapse the separate handling of `calendar_quarter` but we need this because the max area_level might be different for different quarters.

This also explains why it is so slow for new Malawi data, we are working with many quarters there and it is computing it for each slice.

So if we want to make this quicker it is going to require a bit of a bigger change. I'm going to leave this open for now

r-ash:
Actually given this another go now, instead of running each calendar quarter and then binding those rows together, it splits on finest area level. Meaning that if the country has the same finest area level for all calendar quarters it will process these together. Meaning hopefully Malawi runs in reasonable time

R/input-time-series.R
```diff
@@ -88,28 +84,33 @@ aggregate_art <- function(art, shape) {
...
-    dplyr::mutate(year = year_labels(calendar_quarter_to_quarter_id(calendar_quarter)),
-                  quarter = stringr::str_sub(calendar_quarter,-2,-1),
-                  time_period = paste0(year, " ", quarter)) %>%
+  max_levels <- art_number |> dplyr::summarise(area_level = max(area_level),
+                                              .by = "calendar_quarter")
+
+  ## Note we can have the case here that different calendar quarters have
+  ## a different max level e.g. Mozambique
+  art_long <- lapply(unique(max_levels$area_level),
+                     function(level) aggregate_art_by_level(art_number, level)) |>
```
r-ash, line 100: Sorry for noisy PR, this is the important bit of code change

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# naomi 2.9.21
```
jeffeaton, line 1: `DESCRIPTION` has 2.9.20 but specified as 2.9.21 here. Can you confirm which? (Unsure if 2.9.20 is queued up for something else.)

jeffeaton (changes requested):
Thanks for the nice work on this. (And +1 @rtesra for the good tests.)

Only small check to clarify the version number; different in `DESCRIPTION` and `NEWS`.

NEWS.md
r-ash: Thanks, just a typo on my end. I've fixed that in the NEWS

jeffeaton (approved):

rtesra (approved):

PR #429, opened 2024-01-30
Fix issue-122 and add test, add note to README about installation

r-ash:
This PR will
* Fix issue-122, just caused by using a vector where a scalar is expected in an `||`, and tested this
* Also fixed another issue related to comparison report which I saw in the error logs
* Added an installation note into the README


jeffeaton (approved):
Looks good. I made a light clarifying edit on the NEWS entry

PR #433, opened 2024-03-06
Move CI to github actions

r-ash:
In moving this to Avenir we'll no longer have access to the buildkite runners. We can use github actions free runners to achieve what we need instead.

This PR will
* Move CI build to github actions
* Remove branch pin on duckdb (they now have a release which is backwards and forwards compatible, so we can use this to read old results)
* Fix some R CMD check issues
* Fix vignette generation and rerun these


rtesra (approved):

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,59 @@
...
+        shell: bash
+        run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to Docker Hub
```
richfitz, line 26: Do we want to avoid going via docker hub soon and only go via ghcr? You'll need to arrange a new dockerhub account otherwise, and they will definitely consider Avenir commercial

DESCRIPTION
```diff
@@ -74,12 +78,6 @@ LinkingTo:
...
 Remotes:
     bergant/datamodelr,
-    mrc-ide/eppasm,
-    first90=mrc-ide/first90release,
-    reside-ic/traduire,
-    mrc-ide/naomi.options,
-    mrc-ide/mockr,
-    mrc-ide/testthat.buildkite,
-    duckdb/duckdb-r@v0.9.1
+    mrc-ide/naomi.options
```
richfitz, line 81: Is this not on the universe?

NAMESPACE
```diff
@@ -1,5 +1,6 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(naomi::print,naomi_areas)
```
richfitz, line 3: This is odd for two reasons:

* does `print` not work as the generic?
* should naomi not implement this?

R/areas.R
```diff
@@ -160,6 +160,7 @@ create_areas <- function(levels = NULL, hierarchy = NULL, boundaries = NULL,
   v
 }
 
+#' @exportS3Method naomi::print
```
richfitz, line 163: see comments in NAMESPACE

README.md
```diff
@@ -1,6 +1,6 @@
 ## Naomi <img src='images/naomi_hex.png' align="right" height="139" />
 
-[![Travis build status](https://travis-ci.org/mrc-ide/naomi.svg?branch=master)](https://travis-ci.org/mrc-ide/naomi)
```
richfitz, line 3: oh my

docker/teamcity
richfitz, line 1: so much archaeology...

.github/workflows/docker.yaml
r-ash: Yes, expecting we can remove this soon. I'll get deployment working off ghcr and then can remove this

DESCRIPTION
r-ash: It is, removed that line.

NAMESPACE
r-ash: Thanks, think this was me not understanding when you should use `@exportS3method` over `@export`. Have updated this to `@export` now which I think is correct (from https://roxygen2.r-lib.org/articles/namespace.html#s3) I've fixed the signature too to actually match `print`.

.Rbuildignore
```diff
@@ -17,3 +17,6 @@
 ^docker$
 ^buildkite$
 ^images$
+^\.github$
+figure$
+cache$
```
richfitz, line 22:
```suggestion
^figure$
^cache$
```

richfitz (approved):

PR #434, opened 2024-03-20
Use workaround to avoid duckdb raising unnecessary warning

r-ash:
After closing a db connection to duckdb when garbage collection runs we're seeing a warning raised about using `shutdown = TRUE`. We are setting this flag, but the warning is still being raised. We can work around this see https://github.com/duckdb/duckdb-r/issues/60

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 ## Naomi <img src='images/naomi_hex.png' align="right" height="139" />
 
-![R CMD check](https://github.com/mrc-ide/naomi/actions/workflows/R-CMD-check.yml/badge.svg)
+[![Project Status: Active – The project has reached a stable, usable state and is being actively developed.](https://www.repostatus.org/badges/latest/active.svg)](https://www.repostatus.org/#active)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,6 +1,7 @@
 ## Naomi <img src='images/naomi_hex.png' align="right" height="139" />
 
-![R CMD check](https://github.com/mrc-ide/naomi/actions/workflows/R-CMD-check.yml/badge.svg)
+[![Project Status: Active – The project has reached a stable, usable state and is being actively developed.](https://www.repostatus.org/badges/latest/active.svg)](https://www.repostatus.org/#active)
```
r-ash, line 3: Tidying this up whilst I'm here

docker/Dockerfile
```diff
@@ -10,10 +10,8 @@ ENV NAOMI_GIT_BRANCH $GIT_BRANCH
 ENV NAOMI_VERSION $NAOMI_VERSION
 
 RUN install_packages --repo=https://mrc-ide.r-universe.dev \
-    mockr \
     rvest \
-    pkgbuild \
-    testthat.buildkite
```
r-ash, line 16: Tidying these old imports up too

rtesra (approved):

PR #438, opened 2024-07-25
Make duckdb an optional dependency

r-ash:
The PR moves duckdb into suggests, so it should still be installed during tests and development. If you try and save output as a duckdb file without the package installed, it will now error. But you can continue to save naomi output as rds or qs.

Should mean nothing changes on the web but you will no longer be required to install duckdb to install this package (I do need to double check how suggests is working on R universe though)


M-Kusumgar (approved):
LGTM!

PR #443, opened 2024-10-28
Create data comparing aggregated Naomi and national Spectrum programme data

rtesra:
To do:

- [x] Change group in ANC data to females 15-49
- [ ] Update warning message to reference correct name of table /plot
- [x] Add wrapper function to output both ART and ANC and only reads in pjnz and shape file once
- [ ] Add endpoint for warning to be displayed at upload inputs
- [x] Fix tests


R/input-time-series.R
```diff
@@ -565,3 +567,167 @@ build_hierarchy_label <- function(meta_areas) {
...
+  if(!inherits(shape, "sf")) {
+    shape <- read_area_merged(shape) }
+
+  ## Check if art is object or file path
+  if(!inherits(art, c("spec_tbl_df","tbl_df","tbl","data.frame" ))) {
+    art <- read_art_number(art, all_columns = TRUE)}
+
+  ## PJNZ either object or file path
+  if (!inherits(pjnz, "spec_program_data")) {
+    pjnz <- extract_pjnz_program_data(pjnz) }
```
r-ash, line 592: Sorry this is pedantic

```suggestion
    shape <- read_area_merged(shape)
  }

  ## Check if art is object or file path
  if(!inherits(art, c("spec_tbl_df","tbl_df","tbl","data.frame" ))) {
    art <- read_art_number(art, all_columns = TRUE)
  }

  ## PJNZ either object or file path
  if (!inherits(pjnz, "spec_program_data")) {
    pjnz <- extract_pjnz_program_data(pjnz)
  }
```

R/input-time-series.R
```diff
@@ -565,3 +567,167 @@ build_hierarchy_label <- function(meta_areas) {
...
+    spec_aggreagted <- pjnz$art_dec31 |>
+      dplyr::mutate(calendar_quarter = paste0("CY", year, "Q4")) %>%
+      dplyr::count(spectrum_region_code, calendar_quarter, sex, age_group,
+                   wt = art_dec31, name = "value_spectrum")
+  }
+
+  # Get spectrum level to select correct area names
+  spectrum_region_code <- unique(shape$spectrum_region_code)
+
+  if(length(spectrum_region_code) > 1){spectrum_level <- 1}else{spectrum_level <- 0}
```
r-ash, line 619: Could also do

```
spectrum_level <- as.integer(length(spectrum_region_code) > 1))
```

R/input-time-series.R
```diff
@@ -565,3 +567,167 @@ build_hierarchy_label <- function(meta_areas) {
...
+
+##' Compare aggregated subnational Naomi + spectrum totals for comparison table
+##'
+##' @param art Path to file containing ART data or ART data object
+##' @param anc Path to file containing ART data or ART data object
+##' @param shape Path to file containing geojson areas data or areas data object
+##' @param pjnz Path to zip file containing spectrum pjnz file/s
+##' @keywords internal
+
+prepare_spectrum_naomi_comparison <- function(art, anc, shape, pjnz){
```
r-ash, line 707: I think we need to handle the case when either anc or art is missing? Users could upload 1 but not the other.

I'd change signature to
```
prepare_spectrum_naomi_comparison(shape, pjnz, art = NULL, anc = NULL)
```

Then have a check,

```
if (!is.null(art)) {
  if(!inherits(art, c("spec_tbl_df","tbl_df","tbl","data.frame" ))) {
    art <- read_art_number(art, all_columns = TRUE)
  }
  art_comparison <- prepare_art_spectrum_comparison(art, shape, pjnz)
}
```

and similar for ART. Then only need to bind if both present.

r-ash: Can you add a quick test for this too please? You can mock our `prepare_art_spectrum_comparison` and `prepare_anc_spectrum_comparison` to just test that it reads the data and that handles case above if ANC or ART are missing

R/input-time-series.R
```diff
@@ -565,3 +567,167 @@ build_hierarchy_label <- function(meta_areas) {
...
+    art <- read_art_number(art, all_columns = TRUE)}
+
+  ## PJNZ either object or file path
+  if (!inherits(pjnz, "spec_program_data")) {
+    pjnz <- extract_pjnz_program_data(pjnz) }
+
+  art_comparison <- prepare_art_spectrum_comparison(art, shape, pjnz)
+  anc_comparison <- prepare_anc_spectrum_comparison(anc, shape, pjnz)
+
+  rbind(art_comparison, anc_comparison)
```
r-ash, line 730: Possibly worth using `dplyr::bind_rows` here. I think it should generally be faster for large data frames (makes a big difference with a list of data frames) but seems to even with test data here

```
Browse[1]> bench::mark(rbind(art_comparison, anc_comparison), dplyr::bind_rows(art_comparison, anc_comparison))
# A tibble: 2 × 13
  expression           min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result   memory     time       gc      
  <bch:expr>        <bch:> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>   <list>     <list>     <list>  
1 rbind(art_compar… 68.9µs 89.9µs    10047.    4.31KB     4.32  4648     2      463ms <tibble> <Rprofmem> <bench_tm> <tibble>
2 dplyr::bind_rows… 47.5µs 63.3µs    13903.    4.01KB     4.74  5867     2      422ms <tibble> <Rprofmem> <bench_tm> <tibble>
```

r-ash (changes requested):
This looks great, a few comments. Mainly about need to potentially run this with one of anc or art missing.

I also wonder if we should move this out of `input-time-series.R` that is getting quite long and this is data aggregation for a different plot.

R/input-time-series.R
rtesra: Yup much better - thanks!

R/input-time-series.R
rtesra: Have added both and to produce an empt table when user provides no programme data. Kept `rbind` in to be able to bind empty df.

r-ash (approved):

PR #444, opened 2024-10-29
extract metadata without calculating all data

M-Kusumgar:


R/input-time-series.R
```diff
@@ -1,3 +1,153 @@
...
+  ## Check if anc is object or file path
+  if(!inherits(anc, c("spec_tbl_df","tbl_df","tbl","data.frame" ))) {
+    anc <- read_anc_testing(anc)
+  }
+
+  cols_list <- c("anc_clients", "anc_known_pos", "anc_already_art",
+                 "anc_tested", "anc_tested_pos", "anc_known_neg", "births_facility")
+  cols_keep <- intersect(cols_list, colnames(anc))
+  plot_types <- c(cols_keep, "anc_total_pos", "anc_status", "anc_prevalence",
+                  "anc_art_among_known", "anc_art_coverage", "births_clients_ratio")
```
r-ash, line 28: So `cols_list` are the full list of potential indicators in the uploaded data? 

Then `cols_keep` gives us which ones are actually in the data

And `plot_types` is those and any properties derived from these?

What happens if say we don't have `births_facility` in the uploaded data. Then we cannot calculate `births_clients_ratio`? Does that matter that we're returning it in the metadata, or will this not happen?

r-ash (commented):
Looks good on first glance, though have to admit I find `get_art_metadata` a little confusing. Might make more sense with a bit of a closer look!

Would be good to see @rtesra thoughts too as we want to keep maintenance possible for her.

R/input-time-series.R
M-Kusumgar: yep pretty much but `plot_types` doesnt necessarily have to include `cols_keep` like for example in the art metadata, from the code it seemed like there were no if statements to check `births_facility` and it was assumed so i just replicated that

r-ash: Great ok, that all sounds fine! We probably have a validation check somewhere.

r-ash (approved):

PR #450, opened 2024-11-18
fix NA values in survey_hiv_indicators column std_error

jeffeaton:
Address issue flagged about `NA` values in the `std_error` column of `demo_survey_hiv_indicators` where there used to be 0s.  Not sure why it's changed here; but not going to dig into it more unless it comes up again / elsewhere.


jeffeaton (commented):
Thanks — test fixes look fine

r-ash (approved):

PR #453, opened 2024-11-19
Improve population error

rtesra:
Generate more informative warning when user select area level to generate model estimates for which population data is not available


R/model.R
```diff
@@ -329,6 +329,12 @@ naomi_model_frame <- function(area_merged,
   ##           of a Spectrum file and then calibrated. Currently no way to know if areas
   ##           comparise only part of a Spectrum file, so can't address.
 
+  if(any( mf_areas[["area_id"]] %in% unique(population_agesex$area_id) == FALSE)){
```
r-ash, line 332:
```suggestion
  if (!all(mf_areas[["area_id"]] %in% unique(population_agesex$area_id))) {
```

r-ash, line 332:
```suggestion
  if (!all(mf_areas[["area_id"]] %in% unique(population_agesex$area_id))) {
```

R/model.R
```diff
@@ -329,6 +329,12 @@ naomi_model_frame <- function(area_merged,
...
+    stop(
```
r-ash, line 333: Can you add a translation for this error?

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# naomi 2.10.2
+
+* Improve warning generated when user fits model at admin level where no population data exists.
```
r-ash, line 3:
```suggestion
* Improve error generated when user fits model at admin level where no population data exists.
```

r-ash (changes requested):

tests/testthat/test-04-model-frame.R
```diff
@@ -53,7 +53,26 @@ test_that("artnum_mf() works with single quarter ART data", {
...
+                      calendar_quarter4 = "CY2022Q3",
+                      calendar_quarter5 = "CY2023Q3"))
 
+  expect_equal(
+    x$message,
+    paste("Population data not available for admin level selected",
+          "for model projections. Please review model options",
+          "selection to ensure that area level selection is correct.")
+  )
+})
```
jeffeaton, line 75: Could the error message here quote the area level and area level label selected?

> "Population data not available for selected area level 3 (District). Pleas review model options"

Also I'm think I would remove the "for model projections". We usually use "projections" to refer to the T3/T4/T5 projection, but here this area level selection pertains to everything.

R/model.R
```diff
@@ -329,6 +329,10 @@ naomi_model_frame <- function(area_merged,
   ##           of a Spectrum file and then calibrated. Currently no way to know if areas
   ##           comparise only part of a Spectrum file, so can't address.
 
+  if (!all(mf_areas[["area_id"]] %in% unique(population_agesex$area_id))) {
```
jeffeaton, line 332: If the error message is about the area level selected, why is the conditional here referring to the area IDs (not the area level)?

jeffeaton (changes requested):

R/model.R
rtesra: We don't area level available in the population data to easier to check that these intersect and then pull that in if they don't for the error message if needed

PR #456, opened 2024-12-03
[NM-99] Return all indicators for input time series

r-ash:


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# naomi 2.10.3
+
+* Return `anc_already_art`, `anc_status`, `anc_art_among_known` and `anc_total_pos` indicators from ANC input time series data.
```
jeffeaton, line 3: What is `anc_status`? Is that the total denominator?  I usually use `anc_hiv_status` for that. Doesn't matter too much

R/input-time-series.R
```diff
@@ -503,12 +503,7 @@ prepare_input_time_series_anc <- function(anc, shape) {
 
   anc_plot_data_long <- anc_long |>
     dplyr::mutate(!!!mutate_exprs) |>
-    dplyr::select(area_id, area_name, area_level, area_level_label, parent_area_id,
-                  area_sort_order, age_group,  time_period, year, quarter,
-                  calendar_quarter, anc_clients, anc_tested, anc_tested_pos,
-                  anc_prevalence, anc_known_pos, anc_known_neg,
-                  anc_art_coverage, births_facility, births_clients_ratio,
-                  area_hierarchy) |>
+    dplyr::select(-sex) |>
```
jeffeaton, line 506: Is there any risk that `anc_long` has other columns dangling around, or are the columns of that table rigidly specified by an earlier step?

inst/traduire/en-translation.json
```diff
@@ -276,5 +276,12 @@
...
     "NUMBER_ON_ART_DESC": "Number on ART description",
-    "POPULATION_PROPORTION": "Population proportion"
+    "POPULATION_PROPORTION": "Population proportion",
+    "ANC_TOTAL_POS": "Anc total pos",
+    "ANC_TOTAL_POS_DESC": "Anc total pos desc",
+    "ANC_ALREADY_ART_DESC": "ANC already ART desc",
+    "ANC_STATUS": "ANC status",
+    "ANC_STATUS_DESC": "ANC status desc",
+    "ANC_ART_AMONG_KNOWN": "ANC ART among known",
+    "ANC_ART_AMONG_KNOWN_DESC": "ANC ART among known desc"
```
jeffeaton, line 286: "ANC total HIV positive"
"Total HIV positive women at ANC (known positive and tested positive)"
"Women already on lifelong ART before first ANC visit during this pregnancy"
"ANC HIV status ascertained"
"Number of women with HIV status ascertained during ANC (either known or tested at least once)"
"ANC ART coverage among known positive"  _this might be too long?_
"Proportion of women known HIV positive at first ANC visit who were already on ART"

jeffeaton (changes requested):
Thanks, looks good.

See suggestions on the labels. @rtesra can add or suggest others.

inst/traduire/en-translation.json
rtesra: These look good to me

R/input-time-series.R
r-ash: So the indicators are specified by mutate_exprs. So not selecting specific columns here means if we add any more then they should come straight through. Are you worried about some new non-indicator column though?

r-ash: Double checked, no that shouldn't happen. We read the anc file in and specify the columns we want to read. So that should filter out any unwanted stuff. This should just mean all the columns we create during the aggregation get returned automatically

jeffeaton (approved):
Looks good — thanks

PR #460, opened 2024-12-06
[NM-103] Add function to create datapack download standalone

r-ash:


rtesra (approved):

jeffeaton (changes requested):
Per discussion: Make this an Excel download workbook with the data pack CSV on the first sheet, and the metadata discussed on second sheet. (Feel free to split up to more sheets if useful)

What is the filename for the download? Is that specified here or in hintr?

r-ash:
> What is the filename for the download? Is that specified here or in hintr?

That is in hintr. I did mean to ask about that actually but then slipped my mind. At the moment it will be `$iso3_datapack_$date.xlsx` which fits the general structure of the other files. But let me know if you'd like anything else.

jeffeaton:
> > What is the filename for the download? Is that specified here or in hintr?
> 
> That is in hintr. I did mean to ask about that actually but then slipped my mind. At the moment it will be `$iso3_datapack_$date.xlsx` which fits the general structure of the other files. But let me know if you'd like anything else.

Can we do: `$iso3_pepfar-tst-2025_$date-time.xlsx`?

Analogous to we do for main output: `BWA_naomi-output_20241208-2036.zip`

Let's connect tomorrow and see if we can pin down that metadata sheet?

PR #462, opened 2024-12-10
Apply Spectrum ART adjustments

rtesra:
* Update `read_dp_art_dec31()` with new .DP file flags to ensure ART adjustment factor and ART patient reallocation counts are applied to number on ART extracted from Spectrum.
* Ensure adjusted Spectrum number on ART is used in Spectrum-Naomi comparison table.
* Add ART adjustment factor and ART patient reallocation counts to Spectrum-Naomi comparison table.


R/inputs-spectrum.R
```diff
@@ -208,27 +207,73 @@ read_dp_art_dec31 <- function(dp) {
...
+      adult_artadj_absolute <- sapply(dpsub("<AdultPatsAllocToFromOtherRegion>", 3:4, timedat.idx), as.numeric)
+    }
+
     ## Only apply if is number (! is percentage)
     adult_artadj_factor <- adult_artadj_factor ^ as.numeric(!art15plus_isperc)
+    adult_artadj_absolute <- adult_artadj_absolute * as.numeric(!art15plus_isperc)
+
+    ## First add absolute adjustment, then apply scalar adjustment (Spectrum procedure)
+    art15plus_num_adj <- art15plus_num + adult_artadj_absolute
+    art15plus_num_adj <- art15plus_num * adult_artadj_factor
```
jeffeaton, line 252: This adjusted total is being calculated here, but not used later.

What about returning this as additional column `art_dec31_adjusted`, and not needing to return the specific adjustment factors?

R/inputs-spectrum.R
```diff
@@ -208,27 +207,73 @@ read_dp_art_dec31 <- function(dp) {
...
+  art15plus_adjfactor <- as.data.frame.table(adult_artadj_factor,
+                                   responseName = "artadj_factor",
+                                   stringsAsFactors = FALSE)
+  # ART reallocation
+  art15plus_adjabsolute <- as.data.frame.table(adult_artadj_absolute,
+                                             responseName = "artadj_absolute",
+                                             stringsAsFactors = FALSE)
+
+  art15plus$artadj_factor <- art15plus_adjfactor$artadj_factor
+  art15plus$artadj_absolute <- art15plus_adjabsolute$artadj_absolute
```
jeffeaton, line 273: This is probably fine in practice, but relies on the rows of `art15plus_adjfactor` being in the same ordering as `art15plus` (which should be the case based on how they are constructed). But might be slightly more defensive to assign via joining the data frames

R/inputs-spectrum.R
```diff
@@ -284,26 +329,43 @@ read_dp_art_dec31 <- function(dp) {
...
+      child_artadj_absolute <- as.numeric(dpsub("<ChildPatsAllocToFromOtherRegion MV>", 2, timedat.idx))
+    }
+
     ## Only apply if is number (! is percentage)
     child_artadj_factor <- child_artadj_factor ^ !child_art_isperc
+    child_artadj_absolute <- child_artadj_absolute ^ !child_art_isperc
 
+    ## First add absolute adjustment, then apply scalar adjustment (Spectrum procedure)
+    child_art <- child_art + child_artadj_absolute
     child_art <- child_art * child_artadj_factor
```
jeffeaton, line 356: In contrast to above where the adjusted was written into a new variable (`art15plus_num_adj`), here it looks like `child_art` is being overwritten by the adjusted value.

I would have expected these should be consistent?

R/input-comparison.R
```diff
@@ -39,15 +39,26 @@ prepare_art_spectrum_comparison <- function(art, shape, pjnz) {
...
-
+      dplyr::mutate(calendar_quarter = paste0("CY", year, "Q4")) |>
+      dplyr::group_by(spectrum_region_code, calendar_quarter, year, age_group) |>
+      dplyr::summarise(
+        value_spectrum = sum(art_dec31),
+        artadj_factor = sum(artadj_factor),
+        artadj_absolute = sum(artadj_absolute),
+        .groups = "drop") |>
+      dplyr::mutate(
+        value_spectrum_adjusted = (value_spectrum + artadj_absolute) * artadj_factor,
```
jeffeaton, line 50: Per other comments — does this need to be applied here, or can these adjusted values simply be returned by `read_dp_art_dec31()`?

R/input-comparison.R
```diff
@@ -39,15 +39,26 @@ prepare_art_spectrum_comparison <- function(art, shape, pjnz) {
...
+        artadj_factor = dplyr::if_else(age_group == "Y000_014",artadj_factor, artadj_factor/2),
```
jeffeaton, line 51: What is the `/2` in this in this line for?

jeffeaton (commented):
Hi — thanks for this.

Overall comment here: it looks like the function is returning additional columns with the scalar and absolute adjustment factors. Is that required to return both of those, or would it be sufficient to simply return two columns with the raw and adjusted ART totals?

It also looks to me like there is a discrepancy where in the `art_dec31` column might be returning raw numbers for adults and adjusted numbers for children. I might be parsing it wrong though.

tests/testthat/test-input-comparison.R
```diff
@@ -30,6 +30,9 @@ test_that("ART data is properly aggreagted for Spectrum comparison table", {
   expect_equal(unique(x$group), c("art_children", "art_adult_both"))
   expect_equal(unique(x$area_name), c("Malawi - Demo"))
 
+  # TO DO: Ensure ART adjustments in Spectrum are applied
+  # Make a test "pjnz" file
```
r-ash, line 34: This test has been written now right? In `test--spectrum.R`?

r-ash (changes requested):
Some test failures on this, this test

```
 ── Error ('test-input-comparison.R:74:3'): Comparisoon wrapper function works with missing programme data
```

There might some metadata we need to add here too, let's chat this morning.

R/inputs-spectrum.R
rtesra: Updated to output `art_dec31_reported`, `art_dec31_attend` and
`art_dec31_reside`

R/inputs-spectrum.R
rtesra: Updated to be consistent with adult calculations

R/input-comparison.R
rtesra: `read_dp_art31()` updated to output `art_dec31_reported`, `art_dec31_attend` and
`art_dec31_reside`

R/input-comparison.R
rtesra: It was to recalculate the % adjustment factor after aggregating both sexes on ART (edge case for when a user provide adult ART data for all sexes in Naomi) but no longer relevant with change to % adjustment calculated from `art_dec31_reported`, `art_dec31_attend` and
`art_dec31_reside`

R/inputs-spectrum.R
rtesra: Updated to left_join

r-ash (approved):

PR #464, opened 2025-01-02
PEPFAR Datapack 2025 -- national outputs

jeffeaton:
This makes four changes:

* Add national level outputs to PEPFAR Datapack file (filtered from `area_level = 0`)
* Add additional column `country_or_psnu`
* Remove two-year ahead projections from Datapack output (`*.T2` indicators)
* Patch to add the <1 / 1-9 year aggregate for PLHIV_aware_attend indicator 

~Please wait for Christian Domaas to confirm format, then please merge and deploy.~ Christian has reviewed; ready to merge and deploy now.

Will remove the `calendar_quarter_t5` (two-year ahead projection) from the simulation outputs and output zip to reduce the weight of the outputs. Will do this in a separate PR so we don't delay deploying this PEPFAR datapack formatting.

PR #465, opened 2025-01-02
Remove calendar_quarter_t5

jeffeaton:
Remove now unused `calendar_quarter_t5` option and projection step.

Most code is retained but simply commented out.

Merge after #464.

Coincides with UI model options page update: https://github.com/mrc-ide/naomi.options/pull/27

May have implications for hintr — unsure


r-ash (approved):
This looks fine to me, I will say I'm generally not a fan of commenting out code like this, I feel like it makes reading the actual code that is there harder and tends to go stale. So if it was me, I would just remove it, we can always refer back to this PR if we want to add it back in. But up to you if you'd rather keep that in the repo.

PR #468, opened 2025-03-04
Fix build failures on R devel

r-ash:
`sum` function has changed in R devel to error if passed with a vector of `na.rm` so this is now erroring.

Previously `sum(c(1, 2), na.rm = c(TRUE, FALSE))` would just use the first. I've updated this to just use the first explicitly as they should all be the same


M-Kusumgar (approved):
damn thats annoying XD but yh looks good

PR #473, opened 2025-06-29
Calibrate Spectrum Knowledge-of-Status input

jeffeaton:
This implements change to Naomi to ensure the awareness of status calibration aligns to input totals by children / adult male / adult female that are input into the the Spectrum Knowledge of Status editor.

The motivation for this is to capture and reflect the Spectrum calculations for Knowledge of Status among children in Naomi outputs (request from Malawi team).

It ends up being a bit fiddly because the editor is not required to be completed for every year, so some values may need to be interpolated forward / backward.

_Edit 7 July:_ Sorry was going fast last week when I pushed this up. Just a couple more notes: 

- [x] Before merging, I would like to run the updated `extract_pjnz_naomi()` on a large batch of Spectrum files to ensure no unforeseen issues or edge cases
- [x] Would like to ensure @rtesra has reviewed this for awareness of the change


R/calibrate.R
```diff
@@ -606,6 +553,99 @@ calibrate_outputs <- function(output,
...
+#' length as `proportion_raw`.
+#'
+#' @noRd
+calculate_logit_adjustment <- function(proportion_raw,
+                                       denominator_new,
+                                       target_val) {
+    
+  stopifnot(proportion_raw >= 0)
+  stopifnot(proportion_raw <= 1)
+  stopifnot(target_val == target_val[1])
```
r-ash, line 587: How likely are these to error? I'd prefer in general to use an `if (..) { stop("...") }` instead of `stopifnot` because then we can raise a better error message. We should raise translated error messages using `stop(t_("ERROR_KEY"))` if this is going to be surfaced to users.

r-ash (changes requested):
This looks good, nice tidy functions! One question around how we raise errors, but otherwise looks good to me.

R/calibrate.R
jeffeaton: I don't expect these errors to occur in normal use.So not so concerned about about something interpretable to user.


If they do, it indicates something pathological has happened elsewhere in the Spectrum file, or a case that we haven't thought of. 
 
 (This reminds me though; I want to test this on a large batch of Spectrum files before we merge and deploy, and also ensure Rachel has reviewed. Will edit the PR message to note that.)

r-ash (approved):

PR #480, opened 2026-01-19
Replace qs with qs2, make qs a suggested dependency

r-ash:
qs package has been removed from CRAN with no plans to add it back in see https://github.com/qsbase/qs/issues/103. They have an alternative qs2, which inclueds a `qdata` format, but I think that serialization format is not compatible with current qs. We have some data on the deployed web app stored in qs format still, so if we migrate to qs2 we'll need to migrate this data (either to qs2 or duckdb). I'll make a ticket for migrating away from qs.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: naomi
 Title: Naomi Model for Subnational HIV Estimates
-Version: 2.10.16
+Version: 2.10.17
 Authors@R:
     person(given = "Jeff",
            family = "Eaton",
@@ -13,7 +13,7 @@ License: MIT + file LICENSE
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.1
+RoxygenNote: 7.3.3
 Additional_repositories:
     https://mrc-ide.r-universe.dev,
     https://duckdb.r-universe.dev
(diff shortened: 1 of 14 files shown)


jeffeaton (changes requested):
`qs` is a required dependency and requires compiler. Installing it from Github is inconvenient because it means that user needs build tools in order to install Naomi package, even if installing Naomi from `mrc-ide.r-universe.dev` repository.

Would it instead be possible to host compiled versions of `qs` on r-universe repository and install from there?

r-ash:
Yeah we definitely can put it on r-universe. It would be nice to clear the technical debt, but putting on r-universe would give us an interim fix. Package-wise, the changes are fairly minor, it's just the historical data on the server that is the problem. It wouldn't be too bad to migrate that. However we could also move qs into "Suggests" instead of "Imports" and then error if someone tries to save or read a `.qs` file but they don't have the package installed. But have it installed on the server so that could continue to read old files.

jeffeaton:
> Yeah we definitely can put it on r-universe. It would be nice to clear the technical debt, but putting on r-universe would give us an interim fix. Package-wise, the changes are fairly minor, it's just the historical data on the server that is the problem. It wouldn't be too bad to migrate that. However we could also move qs into "Suggests" instead of "Imports" and then error if someone tries to save or read a `.qs` file but they don't have the package installed. But have it installed on the server so that could continue to read old files.

Ah, I see — so, to confirm, the proposal is to:
1. Remove `qs` from current usage in Naomi package
2. Move it to `Suggests` with Github install, for anyone who needs to read an old saved fit object (which would basically just be on the server)

That sounds good to me if that's proposal on table.

jeffeaton (approved):
Thanks for the work on this. This looks good to me.

PR #482, opened 2026-01-21
Add indicator metadata for time series data choropleth

r-ash:
Builds on #480 and should be merged and reviewed after that

This PR adds metadata required for displaying time series plot types in the input review choropleth.


jeffeaton (approved):
I agree this is a bit 'ick'; but fine.

PR #484, opened 2026-03-09
Fix anc areas display t4

rtesra:
* Fix `read_anc_testing()` to replace individual `NA` values in `anc_known_neg` with `0`. Previously only the all-`NA` case was handled; partially-populated
  columns (e.g. data available for recent years only) caused `anc_prev_n` to be `NA` for missing rows, which were then silently dropped, triggering
  a "Column `anc_prev_n` doesn't exist" error in `anc_testing_prev_mf()`.
* Remove requirement for deprecated "display" column in areas file
* Report of mode fro T4 indicators in model fit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ tests/testthat/testdata/fit.RDS
 .vscode
 .Rprofile
 .idea
+.DS_Store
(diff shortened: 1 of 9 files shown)


jeffeaton (approved):
Thanks, this looks fine.
I'm not sure if all of those outputs are needed for t4, but it is surprising that `rho_t4_out` is not outputted, and I can't recall why that might be. Okay to add them all back in.

I fixed small typo in NEWS update

r-ash (approved):

PR #485, opened 2026-06-29
Update warnings to be more helpful to users after 2026 Naomi estimates

rtesra:
* Improve error message when population data area level does not match the
  model `area_level` option. Previously this caused a non-informative
  `stopifnot(!is.na(mf_model[["population_t1"]]))` failure; now the error
  names both the model level and the population data level. (ETH, Debug 43)
* Improve ART duplicate-record error in `artnum_mf()`. Previously the error
  named model-level aggregated areas (which could mislead users to check the
  wrong rows); now it reports the exact input `area_id`, area name, age group,
  sex, quarter, and duplicate count for each offending row. (ETH, Debug 44)
* Add validation in `select_naomi_data()` to detect when ANC already-on-ART
  counts exceed total ANC positives. Previously this produced implausible ART
  coverage values or a non-informative downstream error; now it stops early
  with the affected area, area name, and year. (RWA, Debug 28)


tests/testthat/test-04-model-frame.R
```diff
@@ -30,168 +112,263 @@ test_that("artnum_mf() returns expected number of records", {
...
+  ]
+  art_with_dups <- dplyr::bind_rows(demo_art_number, dup_row)
+
+  expect_error(
+    artnum_mf("CY2017Q4", art_with_dups, a_naomi_mf),
+    "Duplicated records found in ART data for:"
+  )
+  expect_error(
+    artnum_mf("CY2017Q4", art_with_dups, a_naomi_mf),
+    "MWI_4_1_demo (Chitipa) | Y000_014 | sex=both | CY2017Q4 | n=2 records"
```
r-ash, line 147: Nice, I like the `|`, I think they are a bit clearer than what we've tried before when raising errors about rows of data.

r-ash (approved):
This looks good, I'm happy to merge this after we fix the build errors. I'll look into those errors, something failing when building the qs package which is a bit frustrating.

We should definitely check these messages display ok in the app once it is deployed. I think it should be fine, but we should check as they use lots of special characters for formatting the display.

